PR #23, opened 2018-01-15
Implement undo functionality; rollback files that have already been h…

anguss00:
…arvested or stored
Fixes https://github.com/aodn/zzz-aodn-pipeline-poc/issues/99


aodncore/pipeline/files.py
```diff
@@ -221,10 +221,6 @@ def is_stored(self, is_stored):
         self._is_stored = is_stored
         self._post_property_update({'is_stored': is_stored})
 
-    @property
```
lwgordonimos: This could be squashed into the main commit?

anguss00:
Will Squash those commits ;-)

anguss00:
Squashed @lwgordonimos

aodncore/pipeline/files.py
```diff
@@ -83,17 +91,20 @@ def __iter__(self):
         yield 'is_harvested', str(self._is_harvested)
         yield 'is_stored', str(self._is_stored)
         yield 'mime_type', self.mime_type
+        yield 'is_undo_deleted', str(self._is_undo_deleted)
```
lwgordonimos: Maybe we can simplify the language here... "is_undone"

aodncore/pipeline/files.py
```diff
@@ -83,17 +91,20 @@ def __iter__(self):
...
         yield 'name', self.name
         yield 'published', self.published
         yield 'pending_archive', str(self.pending_archive)
         yield 'pending_harvest_addition', str(self.pending_harvest_addition)
         yield 'pending_harvest_deletion', str(self.pending_harvest_deletion)
         yield 'pending_store_addition', str(self.pending_store_addition)
         yield 'pending_store_deletion', str(self.pending_store_deletion)
+        yield 'pending_undo_deletion', str(self.pending_undo_deletion)
```
lwgordonimos: "pending_undo"

aodncore/pipeline/files.py
```diff
@@ -83,17 +91,20 @@ def __iter__(self):
...
         yield 'publish_type', self.publish_type.name
         yield 'should_archive', str(self._should_archive)
         yield 'should_harvest', str(self._should_harvest)
         yield 'should_store', str(self._should_store)
+        yield 'undo_deletion', str(self._undo_deletion)
```
lwgordonimos: "should_undo"

aodncore/pipeline/steps/harvest.py
```diff
@@ -76,7 +76,7 @@ def run(self, pipeline_files):
                 if not deletion_map:
                     self._logger.info('No files to delete, proceeding to add matching files')
                 else:
-                    self.run_deletions(deletion_map, self.tmp_base_dir)
+                    self.run_deletions(deletion_map, self.tmp_base_dir, 'pending_store_deletion')
```
lwgordonimos: Get rid of last attribute since we don't need it anymore once the undo method is totally separate

aodncore/pipeline/steps/harvest.py
```diff
@@ -181,25 +193,36 @@ def execute_talend(self, executor, matched_file_list, talend_base_dir):
...
+            files_to_delete = matched_files.filter_by_bool_attributes_and(deletion_attribute)
+            if files_to_delete:
+                self.upload_runner.run(files_to_delete)
 
-            files_to_delete = matched_files.filter_by_bool_attribute('pending_store_deletion')
+    def run_undo_deletions(self, harvester_map, tmp_base_dir, deletion_attribute):
+        for harvester, matched_files in harvester_map.items():
+            with TemporaryDirectory(prefix='talend_base', dir=tmp_base_dir) as talend_base_dir:
+                for pf in matched_files:
+                    self.remove_symlink(talend_base_dir, pf.dest_path)
```
lwgordonimos: Just replace this with os.remove. It's running out of the temp dir, and the files should definitely exist, so let it throw unhandled, since it would be a very dodgy bug scenario.

aodncore/pipeline/steps/harvest.py
```diff
@@ -181,25 +193,36 @@ def execute_talend(self, executor, matched_file_list, talend_base_dir):
...
+                try:
+                    self.execute_talend(self._config.trigger_config[harvester]['exec'], matched_files, talend_base_dir)
+                except SystemCommandFailedError:
```
lwgordonimos: Get rid of try block, or do some handling... just catching and re-raising...

aodncore/pipeline/steps/harvest.py
```diff
@@ -209,12 +232,13 @@ def run_additions(self, harvester_map, tmp_base_dir):
...
 
-                matched_file_list = [mf.dest_path for mf in matched_files]
-
-                self.execute_talend(self._config.trigger_config[harvester]['exec'], matched_file_list, talend_base_dir)
-
-            for mf in matched_files:
-                mf.is_harvested = True
+                try:
+                    self.execute_talend(self._config.trigger_config[harvester]['exec'], matched_files, talend_base_dir)
+                    for f in matched_files:
```
lwgordonimos: Can use the set_bool_attribute here too.

aodncore/pipeline/steps/harvest.py
```diff
@@ -181,25 +193,36 @@ def execute_talend(self, executor, matched_file_list, talend_base_dir):
...
+                    raise
 
-            for mf in matched_files:
-                mf.is_harvested = True
+            files_to_delete = matched_files.filter_by_bool_attributes_and(deletion_attribute)
+            if files_to_delete:
+                self.upload_runner.run(files_to_delete)
 
-            files_to_delete = matched_files.filter_by_bool_attribute('pending_store_deletion')
+    def run_undo_deletions(self, harvester_map, tmp_base_dir, deletion_attribute):
```
lwgordonimos: This needs the actual temp directory from run_additions, and shouldn't create a new one... replace tmp_base_dir with "additions_base_dir" or something

aodncore/pipeline/steps/upload.py
```diff
@@ -120,6 +121,16 @@ def run(self, pipeline_files):
                                                 dest_path=getattr(pipeline_file, self.dest_path_attr)))
             pipeline_file.is_stored = True
 
+        for pipeline_file in undo_deletions:
+            self._logger.info("undoing upload '{uri}'".format(uri=self._get_absolute_dest_uri(pipeline_file)))
+            try:
+                self._delete_file(pipeline_file)
+                pipeline_file.is_undo_deleted = True
```
lwgordonimos: Move this one out for the try block, to match the other loops

anguss00:
OK, changes up @lwgordonimos

lwgordonimos:
The boss, Travis, ain't happy...

anguss00:
Ready to go @lwgordonimos

aodncore/pipeline/steps/harvest.py
```diff
@@ -176,48 +166,69 @@ def execute_talend(self, executor, matched_file_list, talend_base_dir):
...
+            if files_to_delete:
+                self.upload_runner.run(files_to_delete)
 
-                self.execute_talend(self._config.trigger_config[harvester]['exec'], matched_file_list, talend_base_dir)
+    def run_undo_deletions(self, harvester_map):
+        """Function to un-harvest and undo stored files as appropriate in the case of errors
+        Operates in newly created temporary directory as talend requires a non-existant file to perform un-harvesting
 
-            for mf in matched_files:
-                mf.is_harvested = True
```
lwgordonimos: This is_harvested attribute has been removed from run_deletions, but should be retained. is_harvested is a bit confusingly named, but means "at least one ran on this file", regardless of whether it was an addition or deletion. Given that fact, this attribute should be set in the generic execute_talend method directly after the subprocess is confirmed to have completed without exception.

aodncore/pipeline/steps/upload.py
```diff
@@ -120,6 +121,16 @@ def run(self, pipeline_files):
...
 
+        for pipeline_file in undo_deletions:
+            self._logger.info("undoing upload '{uri}'".format(uri=self._get_absolute_dest_uri(pipeline_file)))
+            try:
+                self._delete_file(pipeline_file)
+            except Exception as e:
+                raise FileDeleteFailedError(
+                    "{e}: '{dest_path}'".format(e=format_exception(e),
+                                                dest_path=getattr(pipeline_file, self.dest_path_attr)))
+            pipeline_file.is_undone = True
```
lwgordonimos: See other comment. Having the attribute here means that a HARVEST_ONLY file would never have this attribute, even after the harvest step ran it through the undo process.

is_undone should be split into two attributes, something like is_storage_undone and is_harvest_undone set at the end of the run_undo_deletions method after each step has been confirmed to have been completed.

aodncore/pipeline/steps/harvest.py
```diff
@@ -176,48 +166,69 @@ def execute_talend(self, executor, matched_file_list, talend_base_dir):
...
-
-                self.execute_talend(self._config.trigger_config[harvester]['exec'], matched_file_list, talend_base_dir)
-
-            for mf in matched_files:
-                mf.is_harvested = True
+                try:
+                    self.execute_talend(self._config.trigger_config[harvester]['exec'], matched_files, talend_base_dir)
+                    matched_files.set_bool_attribute('is_harvested', True)
+                except SystemCommandFailedError:
+                    self.cleanup_on_error()
```
lwgordonimos: Maybe make this method have undo in the name so it's more obvious that it is the key undo entry point?

anguss00:
Changes up Dr. Gordon ;-) @lwgordonimos

anguss00:
@mhidas I reckon this is ready to roll if you want to give it a review

aodncore/pipeline/steps/harvest.py
```diff
@@ -67,75 +66,55 @@ def run(self, pipeline_files):
...
             for event in config_item['events']:
                 for config_type, patterns in event.items():
                     if config_type == 'regex':
                         for pattern in patterns:
                             matched_files_for_pattern = file_list.filter_by_attribute_regex('dest_path', pattern)
+                            match_list = [{'file': f.src_path, 'deletion': f.is_deletion} for f in
+                                          matched_files_for_pattern]
+                            self._logger.sysinfo(
+                                "harvester '{harvester}' matched files: {match_list}".format(harvester=harvester,
+                                                                                             match_list=match_list))
```
mhidas: We need to skip this logging when `match_list` is empty, otherwise it adds a pointless log line for every harvester * every regex.

lwgordonimos:
@mhidas made changes as per previous comments

aodncore/pipeline/steps/harvest.py
```diff
@@ -149,75 +127,100 @@ def executor_conversion(executor):
...
-                matched_file_list = [mf.dest_path for mf in matched_files]
-
-                self.execute_talend(self._config.trigger_config[harvester]['exec'], matched_file_list, talend_base_dir)
-
-            for mf in matched_files:
-                mf.is_harvested = True
+                try:
+                    self.execute_talend(self._config.trigger_config[harvester]['exec'], matched_files, talend_base_dir)
+                except Exception:
+                    self.undo_processed_files()
```
mhidas: I think we need to add the current `{harvester: matched_files}` to the `harvested_file_map` before calling the undo, as there may have been files successfully harvested before the error happened!

In an ideal world one harvester run would be an atomic transaction, but in reality most harvesters commit to the db after each file :-1: :-1:

test_aodncore/pipeline/steps/test_harvest.py
```diff
@@ -147,15 +147,17 @@ def test_multi_harvest_undo(self):
...
+        self.assertTrue(all(f.is_storage_undone for f in harvest_collection))
 
     def test_multi_storage_undo(self):
         harvest_collection = get_storage_collection(True)
         os.environ['PIPELINE_TRIGGER_CONFIG_FILE'] = os.path.join(TEST_ROOT, 'trigger_single_fail.conf')
         multi_harvester_runner = TalendHarvesterRunner(self.uploader, None, TESTDATA_DIR, self.config, self.mock_logger)
         with self.assertRaises(SystemCommandFailedError):
             multi_harvester_runner.run(harvest_collection)
-        self.assertTrue(harvest_collection[0].is_storage_undone)
+        self.assertTrue(all(f.is_storage_undone for f in harvest_collection))
```
mhidas: Couldn't the above three test methods be combined into one? They do all the same setup, right through to running the harvest with an expected error. Just need to combine the assertions at the end. No

mhidas:
@lwgordonimos Thanks, that's better.
Still not undoing the failed harvest itself. Should we or should we not?

test_aodncore/pipeline/steps/test_harvest.py
lwgordonimos: Some are working off a HARVEST_ONLY collection, and some off a HARVEST_UPLOAD collection, so subtly different. I think in general if the setup is cheap, then it's clearer to have separate test methods so it's easier to determine which test actually failed.

mhidas: Ah, I missed that detail. :+1:

mhidas:
@lwgordonimos @anguss00 
I've updated the `published` property, which is used in the notify emails. It should now say 'No' for files that were harvested and/or stored but then undone. See what you think.

lwgordonimos:
@mhidas @anguss00 the 'published' property seems fine to me.

I added the "undo current slice" feature, but it completely and UTTERLY invalidated all the testing assumptions, so I almost had to rewrite that from scratch :/

In a nutshell, I made the test collection out of three files, to test the various combinations of undo behaviour, when slicing and not slicing, and to test how it deals with previously successful (and 'not yet processed' slices).

anguss00:
I reckon this is good to go @lwgordonimos @mhidas. I think we should look at getting this in today so we can test and release (hopefully)

mhidas:
> the 'published' property seems fine to me.
Great, but you seem to have nuked them. I don't see my commit or my changes anywhere :(

mhidas:
I've pushed my commit back up, after updating the change from `is_storage_undone` to `is_upload_undone`
Hope you're still happy with it

aodncore/pipeline/steps/harvest.py
```diff
@@ -218,7 +219,11 @@ def run_additions(self, harvester_map, tmp_base_dir):
                 try:
                     self.execute_talend(self._config.trigger_config[harvester]['exec'], matched_files, talend_base_dir)
                 except Exception:
-                    self.undo_processed_files()
+                    undo_map = [{harvester: matched_files}]
+                    if self.undo_previous_slices:
+                        undo_map.extend(self.harvested_file_map)
```
mhidas: Humph! I just realised this is not quite right... if `undo_previous_slices` is True, it's all good, it will undo everything. But if it's False, it will only undo the failing harvest, _but not any previous harvests on the current slice_!

mhidas: Anyway, the default behaviour (True) is good, so let's go ahead and test that in RC. We can fix up the other option (currently not required) later.

lwgordonimos: Is that as a result of testing? I don't think that's the case, because the harvested_file_map is updated inside the loop, so on a _per-harvester_ basis as soon as it completes successfully. The previous harvests for the current slice should be in the map, and therefore undone. Have you found otherwise from testing?

lwgordonimos: I think you're right that this could be a future enhancement. We could cause the behaviour to differ based on different harvest_params, but should scope that out first.

PR #24, opened 2018-01-15
Dynamic notify fields

lwgordonimos:
This greatly increases the amount of information available to the notify step. Of particular interest is that almost every attribute of the handler can now be included in the messaging templates.


anguss00:
M'kay. Looks good to me!

mhidas:
Yep, that'll do it!

PR #25, opened 2018-01-16
Add more info to notification emails

mhidas:
Tweaks specified in https://github.com/aodn/zzz-aodn-pipeline-poc/issues/152
* incoming directory;
* note to contact info@aodn.org.au for more info;
* celery task name (and id - for AODN reference).

### DIFF
--- a/aodncore/pipeline/templates/notify.html.j2
+++ b/aodncore/pipeline/templates/notify.html.j2
@@ -27,8 +27,8 @@
     }
 </style>
 <body>
-    <p>Dear Uploader,</p>
-    <p>Please find below a report on the handling of the following file:</p>
+    <p>Dear Data Contributor,</p>
+    <p>Please find below a report on the handling of the following file uploaded to the AODN:</p>
     <div class="tablewrapper">
         {{ html_input_file_table }}
     </div>
@@ -40,7 +40,7 @@
 
     {% if html_collection_table %}
     <h4>File collection details:</h4>
-    <p>The input file resolved to the following child file(s):</p>
+    <p>The input file resolved to the following file(s):</p>
     <div class="tablewrapper">
         {{ html_collection_table }}
     </div>
--- a/aodncore/pipeline/templates/notify.html.j2
+++ b/aodncore/pipeline/templates/notify.html.j2
@@ -42,8 +42,20 @@
     </div>
     {% endif %}
 
+    <p>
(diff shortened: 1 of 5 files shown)


aodncore/pipeline/templates/notify.html.j2
```diff
@@ -42,8 +42,20 @@
...
+    (or the relevant AODN project officer).
+    </p>
+
     <p>Regards,<br>
         Australian Ocean Data Network
     </p>
+
+    <footer>
+    <pre>
+    (Log reference: {{ celery_task_name }}[{{ celery_task_id }}])
```
lwgordonimos: I'd just use a "p" tag for this, since then it could be styled to be smaller than the body text, to de-emphasise it a bit. e.g.
```css
    .fineprint {
        font-size: 10px;
    }
```
```html
    <footer>
        <p class="fineprint">
            (Reference: {{ celery_task_name }}[{{ celery_task_id }}])
        </p>
    </footer>
```

aodncore/pipeline/templates/notify.html.j2
```diff
@@ -42,8 +42,20 @@
     </div>
     {% endif %}
 
+    <p>
+    For further information about this message, please contact <a href="mailto:info@aodn.org.au">info@aodn.org.au</a>
+    (or the relevant AODN project officer).
```
lwgordonimos: The "relevant project officer" bit makes me think there should be a per-pipeline list of POs in the notify params that could be used here? Maybe a future enhancement though...

mhidas: Yep, another possible use for a pipeline "owners" list (https://github.com/aodn/zzz-aodn-pipeline-poc/issues/162), but I think the info@aodn will be enough for now.

mhidas:
@lwgordonimos Thanks. I've updated the style for the message footer.

I also updated the note about contacting us. I realised that the emails are sent from info@aodn.org.au, so it's easiest for uploaders to just hit Reply! That way we also get the content of the pipeline notification email (which we may not always get directly).

mhidas:
Why does this PR say checks have failed??
The info on Travis self-contradicting. At the top it says the build was "canceled", but if I scroll down to the bottom of the log, I see:
```
Ran 184 tests in 8.247s
OK
The command "python setup.py test" exited with 0.
Done. Your build exited with 0.
```

mhidas:
I've added a couple more small tweaks, partly just to see if Travis will now give it the :+1:

PR #31, opened 2018-01-22
Add owner related parameters (owner_notify_list, notify_owner_success…

lwgordonimos:
…, notify_owner_error) to the notify_params dict. Assemble recipients into "should_notify" attribute, depending on scenario and notify_params contents.

### DIFF
--- a/aodncore/pipeline/handlerbase.py
+++ b/aodncore/pipeline/handlerbase.py
@@ -205,7 +205,7 @@ class HandlerBase(object):
         self._error_details = None
         self._handler_run = False
         self._instance_working_directory = None
-        self._notify_list = None
+        self._notification_results = None
         self._should_notify = None
 
         self._machine = Machine(model=self, states=HandlerBase.all_states, initial='HANDLER_INITIAL',
@@ -323,12 +323,12 @@ class HandlerBase(object):
         return self._instance_working_directory
 
     @property
-    def notify_list(self):
-        """Read-only property to retrieve the notification list and sent status of each recipient
+    def notification_results(self):
+        """Read-only property to retrieve the notification results, including the sent status of each recipient
 
         :return: NotifyList instance
         """
-        return self._notify_list
+        return self._notification_results
 
     @property
     def result(self):
@@ -538,7 +538,7 @@ class HandlerBase(object):
         self.logger.sysinfo("get_notify_runner -> '{runner}'".format(runner=notify_runner.__class__.__name__))
 
         if self._should_notify:
-            self._notify_list = notify_runner.run(self._should_notify)
+            self._notification_results = notify_runner.run(self._should_notify)
 
     def _notify_success(self):
         self._notify_common()
--- a/aodncore/pipeline/handlerbase.py
+++ b/aodncore/pipeline/handlerbase.py
@@ -206,6 +206,8 @@ class HandlerBase(object):
         self._handler_run = False
         self._instance_working_directory = None
         self._notify_list = None
+        self._should_notify = None
+
         self._machine = Machine(model=self, states=HandlerBase.all_states, initial='HANDLER_INITIAL',
                                 auto_transitions=False, transitions=HandlerBase.all_transitions,
                                 after_state_change='_after_state_change')
@@ -335,6 +337,14 @@ class HandlerBase(object):
         :return: HandlerResult member
         """
         return self._result
+
+    @property
+    def should_notify(self):
+        """Read-only property to retrieve the list of intended recipients *after* being assembled based on notify_params
+
+        :return: NotifyList instance
+        """
+        return self._should_notify
 
     @property
     def start_time(self):
@@ -507,7 +517,7 @@ class HandlerBase(object):
     # 'before' methods for non-ordered state machine transitions
     #
 
-    def _notify_common(self, notify_list_param):
+    def _notify_common(self):
         collection_headers, collection_data = self.file_collection.get_table_data()
         checks = () if self.check_params is None else self.check_params.get('checks', ())
 
@@ -527,17 +537,14 @@ class HandlerBase(object):
         notify_runner = get_notify_runner(notification_data, self.config, self.logger, self.notify_params)
         self.logger.sysinfo("get_notify_runner -> '{runner}'".format(runner=notify_runner.__class__.__name__))
 
-        notify_params_dict = self.notify_params or {}
-        notify_list = notify_params_dict.get(notify_list_param)
-
-        if notify_list:
-            self._notify_list = notify_runner.run(notify_list)
+        if self._should_notify:
+            self._notify_list = notify_runner.run(self._should_notify)
 
     def _notify_success(self):
-        self._notify_common('success_notify_list')
+        self._notify_common()
 
     def _notify_error(self):
-        self._notify_common('error_notify_list')
+        self._notify_common()
 
     def _complete_common(self):
         self.logger.info(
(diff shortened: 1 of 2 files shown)


aodncore/pipeline/handlerbase.py
```diff
@@ -640,24 +647,29 @@ def _handle_error(self, exception, full_traceback=False):
...
-                        'error_notify_list': self.config.pipeline_config['global']['admin_recipients']}
+                if not isinstance(exception, (InvalidConfigError, MissingConfigParameterError)):
+                    should_notify.extend(notify_params_dict.get('owner_notify_list', []))
+
             else:
                 self.logger.error(format_exception(exception))
                 self._error_details = str(exception)
+                should_notify.extend(notify_params_dict.get('error_notify_list', []))
+
+                if notify_params_dict.get('notify_owner_error', False):
```
mhidas: I think the owner should *always* be notified of errors, or at least the default here should be `True`, with the option to turn it off in the config if the owner really wants (... to stick his/her head in the sand... )

aodncore/pipeline/handlerbase.py
```diff
@@ -206,6 +206,8 @@ def __init__(self, input_file,
         self._handler_run = False
         self._instance_working_directory = None
         self._notify_list = None
+        self._should_notify = None
```
mhidas: I think the purpose of these two (`_notify_list` and `_should_notify`) could be made a bit clearer by calling them (`_have_notifed_list` and `_should_notify_list`)

test_aodncore/pipeline/test_dummyHandler.py
```diff
@@ -109,43 +109,124 @@ def test_notify_error(self, mock_smtp):
...
+        self.assertFalse(handler.notify_list[1].notification_succeeded)
+        self.assertIsNone(handler.notify_list[0].error)
+        self.assertIsInstance(handler.notify_list[1].error, InvalidRecipientError)
 
     @mock.patch('aodncore.pipeline.steps.notify.smtplib.SMTP')
     def test_notify_success(self, mock_smtp):
         mock_smtp.return_value.sendmail.return_value = {}
 
         handler = self.run_handler(self.temp_nc_file,
                                    notify_params={'success_notify_list': ['email:nobody1@example.com',
```
mhidas: Could include an `owner_notify_list` here too, and check that it doesn't get included in the recipients.

test_aodncore/pipeline/test_dummyHandler.py
```diff
@@ -109,43 +109,124 @@ def test_notify_error(self, mock_smtp):
         mock_smtp.return_value.sendmail.return_value = {}
 
         handler = self.run_handler_with_exception(ComplianceCheckFailedError, NOT_NETCDF_NC_FILE,
-                                                  notify_params={'error_notify_list': ['email:nobody1@example.com',
-                                                                                       'email:nobody2@example.com']},
+                                                  notify_params={'notify_owner_error': False,
```
mhidas: Ok now I'm _really_ just nitpicking, but the naming of these boolean params just looks a bit wrong to me. Either 'owner_notify_error' or 'error_notify_owner' would look less confusing to me (and similar for the success ones)

lwgordonimos: Nitpicking is relevant here, because it should make sense to a "user".

I think I was going for "notify [the] owner [of] error" and "notify [the] owner [of] success", but happy with whatever makes the most sense.

aodncore/pipeline/handlerbase.py
lwgordonimos: The distinction is between the "system errors" and the "user uploaded the wrong thing and should have known better" processing errors.

System errors always get sent, so this is to have the option to mute some things *if* desired.

I think changing the default to True is my preferred option, to maintain the balance between sensible default but maintaining flexibility.

test_aodncore/pipeline/test_dummyHandler.py
mhidas: Sorry... I agree that makes sense in itself, I'm just comparing it to the other notify_params, but it reallly doesn't matter that much. Never mind.

test_aodncore/pipeline/test_dummyHandler.py
lwgordonimos: Yep, good point. I mostly copied the same dict around for this reason but think I missed this one.

aodncore/pipeline/handlerbase.py
mhidas: Ah right, that makes sense.
We definitely won't want `notify_owner_success` to be be True by default, so maybe we should just leave both defaults as False.

lwgordonimos: I'm not fussed either way, in practical terms, it will just affect what *typically* needs to be in a watch config.

If *most* pipelines want the errors then maybe it makes sense to have True as default, and save having to configure it explicitly in every watch. Or vice versa!

mhidas: Ok, leave both defaults as False. I think in general we all want to receive less email!

aodncore/pipeline/handlerbase.py
```diff
@@ -336,6 +338,14 @@ def result(self):
         """
         return self._result
 
+    @property
+    def should_notify(self):
```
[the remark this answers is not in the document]
lwgordonimos: It's not, but it is just to provide a public property to access the list that was calculated, for example when inspecting the handler object *after* the notify step, and for the string representation.

aodncore/pipeline/handlerbase.py
lwgordonimos: Agreed, this is too ambiguous. I've left "should_upload", as the list of people who *should be* (or should have been) notified, and renamed notify_list to "notification_results", since that is actually what it is... a dict to store the results of the notification attempts.

test_aodncore/pipeline/test_dummyHandler.py
lwgordonimos: Fixed

aodncore/pipeline/handlerbase.py
```diff
@@ -527,17 +537,14 @@ def _notify_common(self, notify_list_param):
...
-        notify_list = notify_params_dict.get(notify_list_param)
-
-        if notify_list:
-            self._notify_list = notify_runner.run(notify_list)
+        if self._should_notify:
+            self._notification_results = notify_runner.run(self._should_notify)
 
     def _notify_success(self):
-        self._notify_common('success_notify_list')
+        self._notify_common()
```
mhidas: Is it still worth having separate `_notify_success` and `_notify_error` methods?

mhidas:
I just one minor question, but otherwise this is good to go!

aodncore/pipeline/handlerbase.py
lwgordonimos: I think so, because they are both associated with a different state transition, so to me it makes sense to be able to potentially act differently for a different transition (even though they are currently just running the same other method!). 

It gives it a 1->1->1 relationship between triggering the state, running it's dedicated method and transitioning to that state, without sharing any "before transition" methods, so keeps it cleaner conceptually.

e.g.
```python
self.state
'HANDLER_POSTPROCESSED'

self.trigger('_trigger_notify_error')  # runs self._notify_error()

self.state
'HANDLER_NOTIFIED_ERROR'
```

mhidas: :+1: Makes sense.

PR #32, opened 2018-01-23
Add schema validation for *_params handler parameters. Consolidate sc…

lwgordonimos:
…hema validation functions to schema module.

Follow on enhancement from: https://github.com/aodn/python-aodncore/pull/31

### DIFF
--- a/aodncore/pipeline/configlib.py
+++ b/aodncore/pipeline/configlib.py
@@ -1,13 +1,12 @@
 import json
 import os
 
-import jsonschema
 from celery import Celery
 from six import iteritems
 
 from .exceptions import InvalidConfigError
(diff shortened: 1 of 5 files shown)


mhidas:
I don't really know how this schema validation thing works, but looks ok to me.

lwgordonimos:
@mhidas it just means those params dicts coming in from the watch config aren't a random bucket of key/values. We can for example ensure that values coming through from JSON which are intended to be booleans actually are booleans.

I've left it a bit open, so that it does allow arbitrary additional values, but it prevents you from supplying an incorrect type to a *known* key.

For example, if you configure 'checks' in 'check_params', it ensures that you set it as an array in the JSON, to pick it up earlier than when it actually gets through to the check step.

PR #34, opened 2018-01-25
Update 'load_json_file' to load into an OrderedDict to maintain the s…

lwgordonimos:
…ame order as the source file. Add some test enhancements to make testing the harvester order simpler and clearer.


mhidas:
Can I merge this so we can test it on RC?

PR #42, opened 2018-02-19
Add some additional examples to HandlerBase documentation for common …

lwgordonimos:
…tasks

I've added some examples for what I could think of as *common* tasks. Feel free to comment, or suggest any additional useful examples, or gaps in documentation (without being too pipeline specific).


aodncore/pipeline/handlerbase.py
```diff
@@ -138,39 +138,123 @@
...
-are typically set from the **params** key of the watch configuration, or from the ``__init__`` method of a handler
-subclass (e.g. when writing tests).
+are typically set from the **params** key of the watch configuration, or as initialisation parameters to ``__init__``
+method of a handler subclass (e.g. when writing tests).
 
 Class parameters
 ~~~~~~~~~~~~~~~~
 
-The following class parameters are also assigned to attributes of the same name, as a convenience.
+The class parameters are also assigned to instance attributes of the same name, as a convenience.
```
mhidas: Why "also"? What else are they assigned to?

aodncore/pipeline/handlerbase.py
```diff
@@ -138,39 +138,123 @@
...
 
-For example, a handler instantiated with any of these class parameters may also access them from the class instance as
-follows::
+A handler instantiated with any of these class parameters may also access them from the class instance::
```
mhidas: For some reason I find the last two statements confusing, even though I'm pretty sure I understand what you're talking about. Above you already explained that "instance attributes" are set from "initialisation parameters" or "params" in the watch config. Here you use the term "class parameters", but I don't think you're saying anything new.
I guess what I'm saying is try to use fewer similar-but-distinct terms.
The examples below certainly help!

aodncore/pipeline/handlerbase.py
```diff
@@ -138,39 +138,123 @@
...
+
+    import os
+    from aodncore.pipeline import PipelineFile
+
+    class MyHandler(HandlerBase):
+        def preprocess(self):
+            # create the product
+            product_path = os.path.join(self.products_dir, 'product.txt')
+            with open(product_path, 'w') as f:
+                f.write('some file contents' + os.linesep)
```
mhidas: Maybe add a comment to say that usually the product would be created from one or more of the files in the collection, which can be accessed via the `self.collection` attribute

lwgordonimos: I was trying to keep each example to one concept, with the overall assumption being that a real pipeline would combine concepts. I had this rolled together with the NetCDF accessing example, but stripped it back to just show how to add a file.

Could make a note in a comment that the product would typically be created from one of the existing files?

aodncore/pipeline/handlerbase.py
lwgordonimos: It's not just you, it's confusing for me to read it back!

I agree, the terminology needs to be simple and consistent, and isn't quite there yet.

What I'm trying to find the terms for are the two distinct concepts of *class arguments/parameters*, i.e. those passed when instantiating the class (the __init__ arguments), and the instance *attributes*. e.g.

```python

class MyHandler(HandlerBase):
    # 'this' and 'that' are class parameters/arguments
    def __init__(self, this, that):
        # 'self.this' and 'self.that' are instance attributes
        self.this = this
        self.that = that

```

I based it on this doco here, which I thought was a nice simple idea, to make the arguments and attributes consistent: http://docs.celeryproject.org/en/latest/userguide/tasks.html#list-of-options

For example, so that a user knows that they can always access the "check_params" they provided to the class via "self.check_params".

Maybe just need to settle on an accepted term for both things, and add a note to explain in simple terms what they mean?

aodncore/pipeline/handlerbase.py
lwgordonimos: Possibly redundant, and related to my other comment, but the also is that the __init__ params are assigned to local variables in the __init__ scope and *also* added as instance attributes.

mhidas: Ok. I don't think the "init scope" is relevant in this doco.

aodncore/pipeline/handlerbase.py
mhidas: Yes. I think all you need to do is explain that _class parameters_ are the arguments of the `__init__` function, which is called when a new handler instance is created. In turn, that function assigns these parameters to corresponding _instance attributes_, which can be accessed within handler methods as `self.<attribute>`. How's that?
The brief example in your reply above might help to make this clear.

aodncore/pipeline/handlerbase.py
```diff
@@ -138,39 +138,123 @@
...
+
+            # create a PipelineFile to represent the product file,
+            # set it's 'publish type' attribute
```
mhidas: it's -> its

aodncore/pipeline/handlerbase.py
```diff
@@ -138,39 +138,123 @@
...
+            # and add it to the handler's file collection
+            product = PipelineFile(product_path)
+            product.publish_type = PipelineFilePublishType.UPLOAD_ONLY
+            self.collection.add(product)
```
bpasquer: i think it is self._file_collection_.add

mhidas:
How's this PR coming along? A couple of conflicts in the docs, but otherwise can we merge it?

lwgordonimos:
Superseded by https://github.com/aodn/python-aodncore/pull/66

PR #43, opened 2018-02-20
Bump compliance checker requirement to latest version

anguss00:
Allows use of origin repo


mhidas:
As discussed, we'll need to use the IMOS plugin from the [new repo](https://github.com/aodn/cc-plugin-imos), so once that's built, we'll also need to update the `cc-plugin-imos` requirement (otherwise it will be incompatible with checker v3).

anguss00:
Does that mean we have to update to cc_plugin_imos==1.1.3 @mhidas?

mhidas:
It will need to be the _next_ version of cc-plugin-imos, i.e. 1.1.4
Or maybe we should call it 1.2.0 so it's a cleaner break from the versions that didn't work with core checker v3.
BTW it would be handy to get the Jenkins build job to also update the version number, as we've done for python-aodncore.

anguss00:
Yep, makes sense. I'll update the jenkins PR as appropriate

requirements.txt
```diff
@@ -1,13 +1,13 @@
 pip
 boto3==1.4.4
-cc-plugin-imos>=1.1.1
+cc-plugin-imos>=1.1.3
 celery==4.0.2
-compliance-checker==2.3.1
+compliance-checker>=3.1.1
```
lwgordonimos: This looks fine to me, assuming we reliably catch version problems/incompatibilities when doing the packaging.

PR #59, opened 2018-03-14
Validate path uniqueness

lwgordonimos:
~~~DON'T MERGE YET, I THOUGHT OF A BETTER WAY~~~


aodncore/pipeline/files.py
```diff
@@ -727,6 +738,37 @@ def set_default_publish_types(self, include_regexes, exclude_regexes, addition_t
             if matches_regexes(f.name, include_regexes, exclude_regexes):
                 f.publish_type = deletion_type if f.is_deletion else addition_type
 
+    def validate_unique_attribute_value(self, attribute, value):
+        """Check that a given attribute value is unique in the collection, and raise an exception if it is already in
+        the collection
+
+        :param attribute: the path to compare with existing paths
+        :param value: the attribute to compare (dest_path/archive_path)
```
mhidas: Given you've gone out of your way to make this a generic "unique attribute" feature, it would be better not to refer to specific attributes (i.e. paths) in the doc string.

mhidas: Also, I think you got them mixed up here. How about this?
```
    :param attribute: the name of the PipelineFile attribute to check
    :param value: the specific attribute value to look for in the collecion
```

aodncore/pipeline/files.py
```diff
@@ -727,6 +738,37 @@ def set_default_publish_types(self, include_regexes, exclude_regexes, addition_t
...
+            raise DuplicateUniqueAttributeError(
+                "{attribute} value '{value}' already set for file(s) '{duplicates}'".format(attribute=attribute,
+                                                                                            value=value,
+                                                                                            duplicates=duplicates))
+
+    def validate_unique_attributes(self, attribute):
+        """Check all paths currently in the collection for uniqueness, and raise an exception if any duplicates are
+            found
+
+        :param attribute: the attribute to compare (dest_path/archive_path)
```
mhidas: Again, make this generic.

lwgordonimos:
@mhidas I've rebased and updated as discussed. If Travis is happy, we should be good to go.

PR #57, opened 2018-03-13
Fixes #55. Only attempt to set 'is_overwrite' for files expected to b…

lwgordonimos:
…e uploaded.

Fixes: https://github.com/aodn/python-aodncore/issues/55

@bpasquer and @lbesnard have independently encountered this bug, related to trying to set "is_overwrite" for files that have no dest_path set, causing the unhandled NoneType exception.

This includes a minor refactor of the upload runner abstract class, in order to make 'set_is_overwrite' a base class method, which calls the abstract, implementation specific '_get_is_overwrite' method in the child classes.

PR #61, opened 2018-03-15
Add retries for S3 file manipulation

### LINKED ISSUE #45: Retry failed S3 uploads
Occasionally S3 uploads fail for no apparent reason (https://github.com/aodn/issues/issues/150). If this happens, wait a bit and retry.

pmbohm:
fixes #45


aodncore/util/external/awsretry.py
```diff
@@ -0,0 +1,136 @@
...
+import logging
+import re
+import time
+from functools import wraps
+
+import boto
+import botocore
+
+__author__ = 'Allen Sanabria'
+__version__ = '1.0.2'
```
mhidas: Should there be a link to the source of this code here?

lwgordonimos: It's already here: https://github.com/aodn/python-aodncore/pull/61/files#diff-c9d6c5c00c596b6bbd4fbc00452260aaR5

I didn't want to modify the source unnecessarily, so we could (in theory, but unlikely) upgrade it in place. It is MIT licensed, so GPLv3 compatible, but if we really had to replace it, not a big deal either.

mhidas: :+1: Ok, that makes sense.

lwgordonimos: Want to make sure we're good open source citizens ;) Also, have fixed the merge conflict, which was caused by https://github.com/aodn/python-aodncore/pull/57

mhidas:
@lwgordonimos Any idea why the Travis tests are failing? It says:
```
AttributeError: 'module' object has no attribute 'test_basestep'
```

lwgordonimos:
Updated to fix tests... by changing library again because the other one was for some reason depending on both boto and boto3...

Issue #46, opened 2018-02-21
Add 'extra_params' capabilities to harvest step


mhidas:
This is to allow handlers to pass additional harvester-specific parameters to Talend jobs.

PR #62, opened 2018-03-19
Harvester extra params

lwgordonimos:
Fixes https://github.com/aodn/python-aodncore/issues/46

This will obviously need some testing in PO box and RC before being accepted for prod.

The key differences are:
* the various "harvester map" dicts have been consolidated into the same structure for consistency
* Similar to the Ruby trigger script, there is now an additional layer in the looping, essentially a "for each event"

It would be nice to revisit with a refactor to get rid of some code duplication (for example between the run_* methods) but I didn't want to change to much at once.

In addition, this obviously makes the undo scenarios more complicated, so we should revisit that at some point, maybe in combination with the refactor I mentioned above. Probably should be a bit less "procedural", which is what I was getting at when introducing the TriggerEvent class and shared map data structure, because the more we want to do with undo behaviour the more structured we will need things like that.

### DIFF
--- a/aodncore/pipeline/steps/harvest.py
+++ b/aodncore/pipeline/steps/harvest.py
@@ -1,3 +1,4 @@
+import itertools
 import os
 import re
 from collections import OrderedDict
@@ -6,11 +7,12 @@ from tempfile import mkstemp
 from .basestep import AbstractCollectionStepRunner
 from ..exceptions import InvalidHarvesterError, UnmappedFilesError
 from ..files import PipelineFileCollection, validate_pipelinefilecollection
-from ...util import mkdir_p, LoggingContext, SystemProcess, TemporaryDirectory
+from ...util import merge_dicts, mkdir_p, LoggingContext, SystemProcess, TemporaryDirectory
 
 __all__ = [
     'get_harvester_runner',
-    'TalendHarvesterRunner'
+    'TalendHarvesterRunner',
+    'TriggerEvent'
 ]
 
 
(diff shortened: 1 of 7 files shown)


aodncore/pipeline/steps/harvest.py
```diff
@@ -72,83 +213,57 @@ def run(self, pipeline_files):
...
-            matched_files = PipelineFileCollection()
+        harvester_map = HarvesterMap()
 
+        for harvester, config_item in self._config.trigger_config.items():
             for event in config_item['events']:
-                for config_type, patterns in event.items():
+                extra_params = None
+                matched_files = PipelineFileCollection()
+
+                for config_type, value in event.items():
```
mhidas: Couldn't we replace this loop with something like this?
```python
    regex_list = event.get('regex', [])
    extra_params = event.get('extra_params')
    for pattern in regex_list:
        matched_files_for_pattern = ...
        ...

    if matched_files:
        event_obj = TriggerEvent(matched_files, extra_params)
```
(Note: I'm pretty sure `event.get()` returns `None` if the given key doesn't exist.)

mhidas:
I've started reviewing but ran out of time. Sorry. Back next week.

I did run a few tests in the PO box, using only harvesters that _don't_ have extra_params, and as far as I could tell everything was still working, including undoing of harvest/store in a multi-harvester scenario.

aodncore/pipeline/steps/harvest.py
lwgordonimos: It's very likely we could replace it, but I didn't really look at that bit because it was in the "just works" category rather than "needs to change to support events" category.

I agree 100% though, the nested loop was basically pulled over directly from the Ruby, so should be refactored into a more Pythonic/less nested style. I'd probably prefer to do it in a follow up PR though, to avoid this one blowing out the scope? What do you reckon?

mhidas: :+1: Sounds good

pmbohm:
@mhidas Are you ready to merge this?

mhidas:
@pmbohm I haven't fully reviewed it yet. I'll look at it soon.

anguss00:
If @mhidas is happy with testing, this looks good from my end @lwgordonimos. Nice one!

anguss00:
Sorry accidentally pressed close and comment

lwgordonimos:
@anguss00 This should be ok.

@bpasquer has tested the behaviour when extra_params is present, and @mhidas I believe tested an existing pipeline for regressions. The test coverage is pretty good for this specifc scenario, so should be fine.

mhidas:
Thanks guys. Sorry I was slow reviewing. Can we test this in RC now?

mhidas:
@bpasquer This is now on RC, if you want to do any more testing before it goes to prod.

Issue #49, opened 2018-02-21
Overwrite status wrong for all S3 files


mhidas:
Overwrite status is meant to indicate when a file being handled already exists at the exact same destination path. Currently it looks like all files going to S3 are reported as overwrites, even if they've never been uploaded before.

PR #65, opened 2018-04-05
Check whether file is present on s3 before flagging overwrite status

anguss00:
Fixes https://github.com/aodn/python-aodncore/issues/49

### DIFF
--- a/aodncore/pipeline/steps/upload.py
+++ b/aodncore/pipeline/steps/upload.py
@@ -221,7 +221,7 @@ class S3UploadRunner(BaseUploadRunner):
     @retry_decorator(**retry_kwargs)
     def _get_is_overwrite(self, pipeline_file, abs_path):
         response = self.s3_client.list_objects_v2(Bucket=self.bucket, Prefix=abs_path)
-        return bool(k for k in response.get('Contents', []) if k['Key'] == abs_path)
+        return bool([k for k in response.get('Contents', []) if k['Key'] == abs_path])
 
     def _post_run_hook(self):
         return
--- a/test_aodncore/pipeline/steps/test_upload.py
+++ b/test_aodncore/pipeline/steps/test_upload.py
@@ -302,6 +302,44 @@ class TestS3UploadRunner(BaseTestCase):
         s3_upload_runner.s3_client.head_bucket.assert_called_with(Bucket=dummy_bucket)
 
     @mock.patch('aodncore.pipeline.steps.upload.boto3')
+    def test_set_is_overwrite_not_present_s3(self, mock_boto3):
+        collection = get_upload_collection()
+        dummy_bucket = str(uuid4())
+        dummy_prefix = str(uuid4())
+        s3_upload_runner = S3UploadRunner(dummy_bucket, dummy_prefix, None, self.test_logger)
+        s3_upload_runner.set_is_overwrite(collection)
+        self.assertEqual(s3_upload_runner.s3_client.list_objects_v2.call_count, 4)
+        self.assertFalse(any(f.is_overwrite for f in collection))
+
+    @mock.patch('aodncore.pipeline.steps.upload.boto3')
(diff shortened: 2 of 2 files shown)


test_aodncore/pipeline/steps/test_upload.py
```diff
@@ -301,6 +301,49 @@ def test_invalid_bucket(self, mock_boto3):
...
+        collection = get_upload_collection()
+        dummy_bucket = str(uuid4())
+        dummy_prefix = str(uuid4())
+        s3_upload_runner = S3UploadRunner(dummy_bucket, dummy_prefix, None, self.test_logger)
+        dest_path = os.path.join('subdirectory', 'targetfile.nc')
+        abs_path = os.path.join(dummy_prefix, dest_path)
+        s3_upload_runner.s3_client.list_objects_v2.return_value = {'Contents': [{'Key': abs_path}]}
+        s3_upload_runner.set_is_overwrite(collection)
+        self.assertEqual(s3_upload_runner.s3_client.list_objects_v2.call_count, 4)
+        for pf in collection:
```
lwgordonimos: I'd condense this to use the filter method on the collection and use assertTrue rather than "equals true". Same on the other tests. 

```python
self.assertTrue(all(f.is_overwrite for f in collection.filter_by_attribute_value('dest_path', dest_path)))
```

P.S. I promise using all() on a generator is legit, unlike bool() :/

test_aodncore/pipeline/steps/test_upload.py
```diff
@@ -301,6 +301,49 @@ def test_invalid_bucket(self, mock_boto3):
...
 
+    @mock.patch('aodncore.pipeline.steps.upload.boto3')
+    def test_set_is_overwrite_not_present_s3(self, mock_boto3):
+        collection = get_upload_collection()
+        dummy_bucket = str(uuid4())
+        dummy_prefix = str(uuid4())
+        s3_upload_runner = S3UploadRunner(dummy_bucket, dummy_prefix, None, self.test_logger)
+        s3_upload_runner.set_is_overwrite(collection)
+        self.assertEqual(s3_upload_runner.s3_client.list_objects_v2.call_count, 4)
+        for pf in collection:
```
lwgordonimos: Similar to below, this can be made more "functional" to kill the loop and multiple assertions:

```python
self.assertFalse(any(f.is_overwrite for f in collection))
```

aodncore/pipeline/steps/upload.py
```diff
@@ -221,7 +221,7 @@ def _get_absolute_dest_uri(self, pipeline_file):
     @retry_decorator(**retry_kwargs)
     def _get_is_overwrite(self, pipeline_file, abs_path):
         response = self.s3_client.list_objects_v2(Bucket=self.bucket, Prefix=abs_path)
-        return bool(k for k in response.get('Contents', []) if k['Key'] == abs_path)
+        return bool([k for k in response.get('Contents', []) if k['Key'] == abs_path])
```
lwgordonimos: Just for reference purposes, the reason this was bugged was because every generator is "True" even if it is "empty", e.g.:

```python
In [1]: l = []

In [2]: g = (e for e in l)

In [3]: g
Out[3]: <generator object <genexpr> at 0x7f752253f410>

In [4]: bool(g)
Out[4]: True

In [5]: g.next()
---------------------------------------------------------------------------
StopIteration                             Traceback (most recent call last)
<ipython-input-5-7dbbdfed0980> in <module>()
----> 1 g.next()

StopIteration:
```

anguss00: Sweet, thanks @lwgordonimos

test_aodncore/pipeline/steps/test_upload.py
anguss00: Cool!

anguss00:
Fixes up @lwgordonimos

PR #66, opened 2018-04-05
Fix docs structure

lwgordonimos:


bpasquer:
greatly improved as compared to the initial version. Good work @lwgordonimos

PR #75, opened 2018-05-02
new quick start page added with storage query example

gsatimos:
could you take a look at this and tell me if you think I should add more while I'm at it?


sphinx/quickstartdoc.rst
```diff
@@ -0,0 +1,25 @@
...
+
+    from aodncore.pipeline import HandlerBase
+    from aodncore.pipeline.config import CONFIG
+
+    class MyHandler(HandlerBase):
+        pass
+
+    h = MyHandler('/path/to/input/file.nc', config=CONFIG, upload_path='/original/incoming/path/file.nc')
+    h.run()
+    sq = handler.state_query
```
lwgordonimos: The primary use case for this will be for the handler itself to run the query from one of the other methods (e.g. preprocess).

To this end, I think it would be good if we could slightly pivot the example to show the typical use case from inside the handler, e.g.

```python
class MyHandler(HandlerBase):
    def preprocess(self):
        results = self.state_query.query_storage('Department_of_Defence/DSTG/slocum_glider/Perth')
        # test the results somehow, usually  to see if there is a "hit" for a particular file
```

The the other scenario you have is more like something a unit test would do, so mention it second. Sound reasonable?

PR #79, opened 2018-05-07
Remove redundant module version related code. Add aodncore version to…

lwgordonimos:
… module versions dict.

### DIFF
--- a/aodncore/pipeline/handlerbase.py
+++ b/aodncore/pipeline/handlerbase.py
@@ -15,10 +15,10 @@ from .files import PipelineFile, PipelineFileCollection
 from .log import SYSINFO, get_pipeline_logger
 from .schema import validate_check_params, validate_harvest_params, validate_notify_params, validate_resolve_params
 from .statequery import StateQuery
-from .steps import (get_cc_module_versions, get_check_runner, get_harvester_runner, get_notify_runner,
-                    get_resolve_runner, get_store_runner)
-from ..util import (format_exception, get_file_checksum, iter_public_attributes, merge_dicts, validate_bool,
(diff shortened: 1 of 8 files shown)


aodncore/pipeline/handlerbase.py
```diff
@@ -429,6 +421,19 @@ def instance_working_directory(self):
         """
         return self._instance_working_directory
 
+    @property
+    def module_versions(self):
+        """Read-only property to access module versions
+
+        :return: module version strings for aodncore, aodndata and compliance checker modules
```
mhidas: I don't see any mention of the aodndata version here. Is it possible to include it?

mhidas:
@lwgordonimos I had a question about including the aodndata version (see above). If that's easy, let's include it, otherwise shall we just merge this?

lwgordonimos:
@mhidas the short answer is that it's not easy to do *correctly*.

I think the one way dependency aodndata <- aodncore is an important boundary to maintain, so any direct imports are a bad idea. 

The other option is to advertise things as an entry point (in the same way as the handlers/dest path functions).

The only other option I could brainstorm would be to add an intermediate base class in the aodndata package which all the handlers inherit from, but I'm less keen on that, especially just for printing a version number!

So I think it might be best to hold off on merging this, until I've had a look at the entry points mechanism. No point merging with the intention to immediately change it anyway!

lwgordonimos:
@mhidas I've updated the PR so that it uses entry points. This will allow us to update the setup.py in aodndata to advertise it's version, and it will be included in the module versions dict automagically.

mhidas:
Nice one, thanks!

Issue #69, opened 2018-04-05
The LazyConfigManager properties should not be mutable


lwgordonimos:
A follow-up to https://github.com/aodn/python-aodncore/issues/67.

The LazyConfigManager properties are generally "dict" types, which makes it too easy to get a reference to a value and accidentally modify it. This is a big problem because the way the tasks are set up currently, the CONFIG object is created once and then used repeatedly by unrelated task executions, which means accidentally modifying it corrupts it for all subsequent task executions, e.g.

1. file A arrives and is handled. The handler instance inadvertently modifies the config.
1. file B arrives and is handled. The config was corrupted by the file A handler, and this further corrupts it
1. file C arrives and the handler reuses the corrupt config (and probably corrupts it more!)

PR #80, opened 2018-05-14
Resolve step fixes

lwgordonimos:
Fix bug with manifest file handling by not popping the value out of the dictionary. Add a couple of optimisations for large rsync manifest files.

This overlaps with the following issue regarding config mutability: https://github.com/aodn/python-aodncore/issues/69


anguss00:
Will merge as trav failures are unrelated

PR #83, opened 2018-05-21
Upgrade to compliance checker 4

mhidas:
~~DO NOT MERGE until the chef code used to pin the checker version has been removed.~~

This simply updates the requirements file to specify the latest version of the compliance checker, and the imos plugin.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,8 +1,8 @@
 pip
 boto3==1.4.4
-cc-plugin-imos>=1.2.0
+cc-plugin-imos>=1.3.0
 celery==4.0.2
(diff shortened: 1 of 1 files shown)


mhidas:
Also need to deploy cc-plugin-imos 1.3.0 to prod before this can be deployed. That's why the Travis checks failed.

mhidas:
@lwgordonimos @anguss00 Can we merge this now so we can test on RC?

I haven't promoted the plugin 1.3.0 to prod yet, but will do as soon as I've confirmed that it's all working ok in RC.

PR #86, opened 2018-05-25
PipelineFileCollection enhancements

lwgordonimos:
Makes it easier to manipulate/filter the collection in some useful ways. Primarily to avoid unnecessary "for" loops in child handlers which are a common source of bugs when trying to do basic things like set attributes on files.

### DIFF
--- a/aodncore/pipeline/files.py
+++ b/aodncore/pipeline/files.py
@@ -8,8 +8,9 @@ import six
 from .common import (FileType, PipelineFilePublishType, PipelineFileCheckType, validate_addition_publishtype,
                      validate_checkresult, validate_checktype, validate_deletion_publishtype, validate_publishtype)
 from .exceptions import DuplicateUniqueAttributeError, DuplicatePipelineFileError, MissingFileError
+from .schema import validate_check_params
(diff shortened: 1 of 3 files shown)


anguss00:
Beautiful

aodncore/pipeline/files.py
```diff
@@ -788,7 +806,32 @@ def set_file_update_callback(self, file_update_callback):
...
+        """
+        if check_params is None:
+            check_params = {}
+        else:
+            validate_check_params(check_params)
+
+        checks = check_params.get('checks', ())
+
+        all_deletions = PipelineFileCollection(f for f in self.__s if f.is_deletion)
+        all_additions = PipelineFileCollection(self.__s.difference(all_deletions))
```
mhidas: Could this just be `self.difference(all_deletions)`?

lwgordonimos: It probably *could* be but the reasoning is that I've been trying to avoid public methods which loop back and call other public methods on the same instance. Everything else operates against the internal set, so I didn't want to change this convention just for this one.

It could potentially be slackened off overall to use the public methods from the same instance(for example, to reduce duplication of similar operations against the set in multiple places), but I'd want that to be a conscious choice for everything at once I think (and in a different PR).

PR #89, opened 2018-05-30
Handle invalid handler params

lwgordonimos:
~~~DO NOT MERGE YET - found some issues in PO box~~~

### DIFF
--- a/aodncore/pipeline/watch.py
+++ b/aodncore/pipeline/watch.py
@@ -19,8 +19,8 @@ import abc
 import logging.config
 import os
 import stat
+import warnings
 from uuid import uuid4
-from warnings import warn
 
 from six import PY2
 from transitions import Machine
@@ -49,6 +49,11 @@ from six import iteritems
 
 from .exceptions import InvalidHandlerError
 from ..util import list_regular_files, safe_move_file
+
+# Filter noisy and useless numpy warnings
+# Reference: https://github.com/numpy/numpy/pull/432
+warnings.filterwarnings("ignore", message="numpy.dtype size changed")
+warnings.filterwarnings("ignore", message="numpy.ufunc size changed")
 
 __all__ = [
     'get_task_name',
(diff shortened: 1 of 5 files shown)


lwgordonimos:
@mhidas this should be good to review now. I discovered that raising an Exception wasn't really what we wanted since it breaks the loop of tasks, causing a single misconfigured pipeline to break *all* of the pipelines.

This solution raises a warning in the Celery log and continues in the loop, so that only misconfigured pipelines will be unavailable (i.e. leave the files in incoming).

There are a couple of other necessary watch related updates as well, which I've separated by commits but put in the same PR, since it is under the not too broad umbrella of "fix watchservice, including correcting logging'.

aodncore/pipeline/watch.py
```diff
@@ -211,8 +221,15 @@ def _register_tasks(self):
...
 
             params = items.get('params', {})
-            task_object = self._build_task(pipeline_name, handler_class, params)
-            self._application.register_task(task_object)
+
+            try:
+                _ = handler_class('', config=self._config, **params)
+            except TypeError as e:
+                warnings.warn("invalid parameters for handler '{name}': {e}".format(name=items['handler'],
+                                                                                    e=format_exception(e)))
```
mhidas: This warning appears in the watchservice log, and in every task's stderr.log, and there's currently no indication which one has the bad config. Could we add the task name in the message?

Better still, is it possible to have this appear only in the badly configured task's log?

lwgordonimos: > This warning appears in the watchservice log, and in every task's stderr.log, and there's currently no indication which one has the bad config. Could we add the task name in the message?

Yes, this is an oversight, fixing.

> Better still, is it possible to have this appear only in the badly configured task's log?

Not without a deeper refactor of the way the workers work.

The way it is, each worker is *aware* of the other tasks, but only picks up tasks that it is supposed to. So each task *sees* the errors but they aren't actually a problem for that pipeline. Given that it will be picked up in RC, I think we can leave this enhancement until later!

mhidas: :+1:

PR #91, opened 2018-05-31
Safe default publishtype to avoid /dev/null handlers

lwgordonimos:
DO NOT MERGE YET, for discussion.

1. files start as UNSET
1. the include/exclude mechanism for the collection will change matching files from UNSET->SOMETHING
1. anything remaining as UNSET by the time it gets to the publish step will cause an error

This means that the handler and/or the include regexes are responsible for assigning a type to each file (which could well be NO_ACTION, but that would be an explicit, deliberate affirmative action, e.g. if there are excluded files, you'd need something in preprocess/process to set their publish types to an appropriate value).

Note this is still distinct from the 'allowed_regexes' change, which only deals with the input filename matching. This is purely for assigning publish types to files in the collection *after* the input file has been deemed to be an acceptable name by the handler.


aodncore/pipeline/common.py
```diff
@@ -169,6 +169,7 @@ class PipelineFilePublishType(Enum):
     __slots__ = ('_is_addition_type', '_is_deletion_type', '_is_archive_type', '_is_store_type', '_is_harvest_type')
 
     # valid for both addition and deletion types
+    UNSET = (True, True, None, None, None)
```
mhidas: I think this should be `UNSET = (None, None, None, None, None)` I've tried that and all the unittests still pass. 

If some code tried to set a publish_type to UNSET using the `PipelineFile.publish_type` setter, it would then fail the validation, and so it should. UNSET should only ever be an initial state.

test_aodncore/pipeline/test_dummyHandler.py
```diff
@@ -43,14 +44,17 @@ def test_dest_path_from_string(self):
         self.assertIs(handler._dest_path_function_ref, dest_path_testing)
 
     def test_include(self):
-        handler = self.run_handler(BAD_ZIP, include_regexes=['good\.nc'])
-        eligible_filenames = [f.name for f in handler.file_collection if f.should_harvest]
+        handler = self.run_handler_with_exception(UnmatchedFilesError, BAD_ZIP, include_regexes=['good\.nc'])
+        eligible_filenames = handler.file_collection.filter_by_attribute_id_not('publish_type',
+                                                                                PipelineFilePublishType.UNSET).get_attribute_list('name')
         self.assertIn('good.nc', eligible_filenames)
         self.assertNotIn('bad.nc', eligible_filenames)
```
mhidas: I know it's not part of this PR, but this would be a stronger assertion than the above two lines:
```python
self.assertListEqual(['good.nc'], eligible_filenames)
```
Same in `test_exclude` below.

test_aodncore/pipeline/test_storage.py
```diff
@@ -225,7 +225,7 @@ def test_set_is_overwrite_with_no_action_file(self):
 
         temp_file = PipelineFile(self.temp_nc_file)
         self.assertIsNone(temp_file.dest_path)
-        self.assertIs(temp_file.publish_type, PipelineFilePublishType.NO_ACTION)
+        self.assertIs(temp_file.publish_type, PipelineFilePublishType.UNSET)
```
mhidas: Update test method name too?
Is this the same test as `test_set_is_overwrite_with_unset_file` in pipeline/steps/test_store.py ?

aodncore/pipeline/common.py
lwgordonimos: Yes, good idea. The key thing was to make sure the tuple was different from everything else, hence the 'None's, but it makes perfect sense to extend it out to all Nones now that I look at it.

test_aodncore/pipeline/test_dummyHandler.py
lwgordonimos: Those have been there since the start I reckon! But yes, it should be updated.

test_aodncore/pipeline/test_storage.py
lwgordonimos: This one tests directly against a broker object, the other tests setting through the runner object. Looking at it, it's largely the same, but I think this was more of an internal test, whereas the other is the interface that the handler itself uses.

lwgordonimos:
@mhidas I've made those changes, also added explicit tests for trying to set a file to UNSET.

PR #95, opened 2018-06-05
add option to parse an attribute as a timestamp in FileClassifier

mhidas:
This was done in data-services (https://github.com/aodn/data-services/pull/740/commits/94c5563115cd063a54221fff0794be8e901567be) after we moved FileClassifier into aodncore.

It is needed for ABOS_SOTS pipeline2.

### DIFF
--- a/aodncore/pipeline/fileclassifier.py
+++ b/aodncore/pipeline/fileclassifier.py
@@ -24,6 +24,8 @@ Expected use::
 import os
 import re
 
+from datetime import datetime
+
 from netCDF4 import Dataset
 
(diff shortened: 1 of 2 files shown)


mhidas:
Thanks @lwgordonimos . Could we please merge this so I can get on with the ABOS_SOTS pipeline?

mhidas:
@bpasquer @lbesnard FYI, some new functionality that will be available in all FileClassifier objects when this gets merged. In case it's useful.

mhidas:
:+1: Thanks!

PR #96, opened 2018-06-06
Add ablility to validate dest_path/archive_path before the handler performs any destructive actions

lwgordonimos:
Following on from: https://github.com/aodn/python-aodndata/pull/46

This allows handlers to be configured to exit with an error before publishing anything with an invalid dest_path, as determined by the handler parameters.

For example, Argo would be configured as follows:

```python
handler = ArgoHandler(input_file, allowed_dest_path_regexes=['IMOS/Argo'])
```

Then if, for any reason(e.g. a bug) manages to generate a dest_path outside of 'IMOS/Argo', the handler will error out before publishing anything (i.e. harvesters/uploads/deletions only happen if *every* file matches one of the given regexes).

### DIFF
--- a/aodncore/pipeline/files.py
+++ b/aodncore/pipeline/files.py
@@ -7,7 +7,7 @@ import six
 
 from .common import (FileType, PipelineFilePublishType, PipelineFileCheckType, validate_addition_publishtype,
                      validate_checkresult, validate_checktype, validate_deletion_publishtype, validate_publishtype)
-from .exceptions import DuplicateUniqueAttributeError, DuplicatePipelineFileError, MissingFileError
+from .exceptions import AttributeValidationError, DuplicatePipelineFileError, MissingFileError
 from .schema import validate_check_params
 from ..util import (IndexedSet, format_exception, get_file_checksum, iter_public_attributes, matches_regexes,
                     slice_sequence, validate_bool, validate_callable, validate_mapping,
@@ -866,7 +866,7 @@ class PipelineFileCollection(MutableSet):
         """
         duplicates = [f for f in self.__s if getattr(f, attribute) == value]
         if duplicates:
-            raise DuplicateUniqueAttributeError(
+            raise AttributeValidationError(
                 "{attribute} value '{value}' already set for file(s) '{duplicates}'".format(attribute=attribute,
                                                                                             value=value,
                                                                                             duplicates=duplicates))
@@ -887,7 +887,7 @@ class PipelineFileCollection(MutableSet):
             duplicates = []
             for value in duplicate_values:
                 duplicates.extend(f for f in self.__s if getattr(f, attribute) == value)
-            raise DuplicateUniqueAttributeError(
+            raise AttributeValidationError(
                 "duplicate paths found for files '{duplicates}'".format(duplicates=duplicates))
 
 
--- a/aodncore/pipeline/files.py
+++ b/aodncore/pipeline/files.py
@@ -871,6 +871,20 @@ class PipelineFileCollection(MutableSet):
                                                                                             value=value,
                                                                                             duplicates=duplicates))
 
+    def validate_attribute_value_matches_regexes(self, attribute, include_regexes):
+        """Check that the given :py:class:`PipelineFile` attribute matches at least one of the given regexes
+
+        :param attribute: the attribute to compare
+        :param include_regexes: list of regexes of which the attribute must match at least one
+        :return: None
+        """
+        unmatched = [f for f in self.__s if
+                     not matches_regexes(getattr(f, attribute), include_regexes=include_regexes)]
+        if unmatched:
+            raise AttributeValidationError(
+                "invalid attribute values found for files: {unmatched}. Must match one of: {regexes}".format(
+                    unmatched=unmatched, regexes=include_regexes))
+
     def validate_attribute_uniqueness(self, attribute):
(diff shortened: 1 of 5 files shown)


bpasquer:
Just wondering whether in the case of PipelineFile deletion the validity check should be required more than being an option. Cause, correct me if I am wrong, but if the _allowed_dest_path_regexes_ is not set, validity checks will not be performed and the handler can generate  invalid paths. Is that right?

lwgordonimos:
That is correct. It's optional at the configuration level and not in handler code. BUT if you configure the pipeline with this constraint, then it *is* mandatory whenever the handler runs.

bpasquer:
Yes, and that is my point. It's mandatory if we decide it should be (by configuring the pipeline with this constraint), but shouldn't it be mandatory for deletions? The deletion would only be allowed _if_ the constraint exists.

lwgordonimos:
Can do, but would need some thinking to do it in a sensible way, so as to avoid bloat and adding too much paranoid validation in weird places ;)

bpasquer:
👍

aodncore/pipeline/files.py
```diff
@@ -863,6 +863,20 @@ def validate_unique_attribute_value(self, attribute, value):
                                                                                             value=value,
                                                                                             duplicates=duplicates))
 
+    def validate_attribute_value_matches_regexes(self, attribute, include_regexes):
+        """Check that the given :py:class:`PipelineFile` attribute matches at least one of the given regexes
```
mhidas: ... for _every_ file in the collection.

aodncore/pipeline/files.py
```diff
@@ -863,6 +863,20 @@ def validate_unique_attribute_value(self, attribute, value):
...
+
+        :param attribute: the attribute to compare
+        :param include_regexes: list of regexes of which the attribute must match at least one
+        :return: None
+        """
+        unmatched = [f for f in self.__s if
+                     not matches_regexes(getattr(f, attribute), include_regexes=include_regexes)]
```
mhidas: I think the error message would be a bit more legible if this list only included the file name and the attribute in question. Maybe make it a dict?
```python
    unmatched = {f.name: getattr(f, attribute)
￼                 for f in self.__s
￼                 if not matches_regexes(getattr(f, attribute), include_regexes=include_regexes)}

```

aodncore/pipeline/files.py
```diff
@@ -863,6 +863,20 @@ def validate_unique_attribute_value(self, attribute, value):
...
+        if unmatched:
+            raise AttributeValidationError(
+                "invalid attribute values found for files: {unmatched}. Must match one of: {regexes}".format(
+                    unmatched=unmatched, regexes=include_regexes))
```
mhidas: The error message should include the name of the attribute.

aodncore/pipeline/handlerbase.py
```diff
@@ -659,16 +673,25 @@ def _store_unharvested(self):
...
 
-    def _publish(self):
+    def _pre_publish(self):
         self.file_collection.set_archive_paths(self._archive_path_function_ref)
         self.file_collection.validate_attribute_uniqueness('archive_path')
-        self._archive()
+
+        if self.allowed_archive_path_regexes:
+            self.file_collection.validate_attribute_value_matches_regexes('archive_path',
+                                                                          self.allowed_archive_path_regexes)
```
mhidas: Shouldn't this be applied only to files that will actually be archived?
Likewise for the dest_path validation, should only be applied to files that will be stored, harvested or deleted.

mhidas:
Regarding @bpasquer's comment, could we make `allowed_dest_path_regexes` a mandatory config item for each pipeline?

aodncore/pipeline/handlerbase.py
lwgordonimos: Yes it should!

lwgordonimos:
Re mandatory parameter: I think it's reasonable to make it mandatory *at some point*, but the question is "where is best". Just off the top of my head:

1. adding to the handler means every single unit test now needs to make sure it's adding the attribute.
1. adding to the StoreRunner means you could have a concept of a read only store runner UNLESS the patterns were set. This means the handler class doesn't have to deal with it (separating concerns is good...) and you could do things like the StateQuery class not having access to any deletion methods at all, e.g. it would get rid of the risk of this sort of thing from a handler which breaks the "don't touch underscored methods" convention and runs things like "self.state_query._broker.delete(myfiles)"
1. adding it to the watch configuration layer would mean that test etc. can be less concerned with it (unless they explicitly want to test it), but when it gets *deployed* somewhere, the task creation could be rejected if it didn't have the mandatory param(s)

lwgordonimos:
@mhidas updated with those suggestions, all made good sense

mhidas:
@lwgordonimos I was thinking option 3, i.e. adding allowed_dest_path_regexes as a required parameter in the watch config json schema.

mhidas:
@lwgordonimos This needs rebasing now.

lwgordonimos:
@mhidas fixed merge conflicts

mhidas:
Thanks @lwgordonimos . There are a couple of other PRs that have been merged since the last build (in particular #91). Might be worth doing a build with just those changes first, once Jenkins allows, before merging this too?

lwgordonimos:
Yes sounds good to me. Don't want too many major-ish changes going in at the same time!

PR #98, opened 2018-06-06
Update LocalFileStorageBroker to use the same sort key as 'list_regul…

lwgordonimos:
…ar_files'.

This was an omission in the LocalFileStorageBroker which I noticed when reviewing https://github.com/aodn/python-aodndata/pull/46. There's no point accessing the results by numeric index unless it's in a predictable order!

EDIT: ~~actually, DO NOT MERGE YET~~. 

We should replicate the sorting used by S3, "List results are always returned in UTF-8 binary order.".

https://docs.aws.amazon.com/AmazonS3/latest/dev/ListingKeysUsingAPIs.html

EDIT2: I have now updated to test the specific order of results to match S3 behaviour


mhidas (approved):
Seems to work ok (i.e. doesn't break unittests here or in aodndata)

Issue #97, opened 2018-06-06
Specify additional check params separately for each check suite


mhidas:
Currently we're only using the `check_params` config element to specify which check suites (cf / imos) to apply. However, the checker accepts other parameters (see [ComplianceChecker.run_checker](https://github.com/ioos/compliance-checker/blob/master/compliance_checker/runner.py#L34)) that control its behaviour. Some of these (in particular `skip_checks` and/or `criteria`) could be used to circumvent some troublesome checks in specific cases (see https://github.com/aodn/internal-discussions/issues/410). However, these parameters may need to be different for CF and IMOS checks.

The current config only allows one set of parameters, e.g. 
```
  "check_params": {
    "checks": [ "cf", "imos:1.4" ],
    "skip_checks": "check_dimension_order"
  }
```

Instead, we could have
```
  "check_params": {
    "cf": {
      "criteria": "normal"
      "skip_checks": "check_dimension_order"
    },
    "imos:1.4": {
      "criteria": "strict"
    }
  }
```

mhidas:
I just realised the `skip_checks` option expects a list, so I've updated the examples above.

PR #99, opened 2018-06-07
Add skip_checks option to check_params

mhidas:
This is needed right now in order to allow skipping a few annoying checks in some pipelines.

It's related to #97, but is a smaller, simpler change that doesn't require updating all the existing watch configs.

### DIFF
--- a/aodncore/pipeline/schema.py
+++ b/aodncore/pipeline/schema.py
@@ -19,6 +19,7 @@ CHECK_PARAMS_SCHEMA = {
     'properties': {
         'checks': {'type': 'array'},
         'criteria': {'type': 'string'},
+        'skip_checks': {'type': 'array', 'items': {'type': 'string'}},
         'output_format': {'type': 'string'},
         'verbosity': {'type': 'integer'}
     },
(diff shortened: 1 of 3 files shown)


lwgordonimos:
Looks good to me. 

I was going to comment on the confusing nature of checks vs. skip checks vs. tests vs. suites but that seems to be the upstream terminology?

Having checks and skip_checks in the same set of params but meaning different things is very confusing!

lwgordonimos:
```
  --test TEST, -t TEST, --test= TEST, -t= TEST
                        Select the Checks you want to perform. Defaults to
                        'acdd' if unspecified. Versions of standards can be
                        specified via `-t <test_standard>:<version>`. If
                        `<version>` is omitted, or is "latest", the latest
                        version of the test standard is used.

  --skip-checks SKIP_CHECKS, -s SKIP_CHECKS
                        Specifies tests to skip
```

Case in point... the tests option "tests" selects "checks" to perform, but the "skip checks" option specifies "tests to skip"... wut?

mhidas:
Yes, confusing, I agree, but as you say it's just the terminology from IOOS. 

We could try to be more specific in our config and have options "check_suites" (instead of the current "checks"), and "skip_methods" (instead of "skip_checks"), but internally we'd still have to map these to the parameters expected by the core checker. I don't think it's worth the effort.

Also, if we later refactor the whole check_params group as I suggested in #97, then we won't explicitly have a "checks" option. The suite names will become dict keys.

PR #100, opened 2018-06-07
Utils - add function to retrieve fields from a filename given a compi…

lbesnard:
…led pattern

### DIFF
--- a/aodncore/util/misc.py
+++ b/aodncore/util/misc.py
@@ -21,6 +21,7 @@ __all__ = [
     'discover_entry_points',
     'ensure_writeonceordereddict',
     'format_exception',
+    'get_pattern_subgroups_from_string',
     'is_nonstring_iterable',
     'is_function',
     'is_valid_email_address',
@@ -93,6 +94,22 @@ def format_exception(exception):
     :return: string
     """
     return "{cls}: {message}".format(cls=exception.__class__.__name__, message=exception)
+
+
+def get_pattern_subgroups_from_string(string, pattern):
+    """Function to retrieve parts of a string given a compiled pattern (re.compile(str))
+    :return: dictionary of fields matching a given pattern
+    """
+    retype = type(re.compile(''))
+    if not isinstance(pattern, retype):
+        try:
+            pattern = re.compile(pattern)
+        except Exception as err:
+            raise TypeError("error compiling pattern '{pattern}': {e}".
+                            format(pattern=pattern, e=format_exception(err)))
+
+    m = pattern.match(string)
+    return {} if m is None else m.groupdict()
 
 
 def is_function(o):
--- a/aodncore/util/misc.py
+++ b/aodncore/util/misc.py
@@ -97,7 +97,13 @@ def format_exception(exception):
 
 
 def get_pattern_subgroups_from_string(string, pattern):
-    """Function to retrieve parts of a string given a compiled pattern (re.compile(str))
+    """Function to retrieve parts of a string given a compiled pattern (re.compile(pattern))
+    the pattern needs to match the beginning of the string
+    (see https://docs.python.org/2/library/re.html#re.RegexObject.match)
+
+    * No need to start the pattern with "^"; and
+    * To match anywhere in the string, start the pattern with ".*".
+
     :return: dictionary of fields matching a given pattern
     """
     retype = type(re.compile(''))
--- a/test_aodncore/util/test_misc.py
+++ b/test_aodncore/util/test_misc.py
@@ -1,4 +1,5 @@
 import os
+import re
 import sys
 import uuid
 from collections import OrderedDict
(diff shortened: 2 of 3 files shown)


aodncore/util/misc.py
```diff
@@ -95,6 +96,24 @@ def format_exception(exception):
...
+def get_fields_from_filename(filepath, pattern):
+    """Function to retrieve parts of a filename given a compiled pattern (re.compile(str))
+    :return: dictionary of fields matching a given pattern
+    """
+    retype = type(re.compile(''))
+    if not isinstance(pattern, retype):
+        try:
+            pattern = re.compile(pattern)
+        except Exception as err:
+            raise TypeError("pattern: \"{pattern}\" could not be compiled as a \"{retype}\" instance".format(
```
lwgordonimos: I don't really like escaping quotes, so usually use single quotes *inside* the string. Also, I think the message can be simpler, since the type stuff isn't really relevant to know. What is relevant is what the internal error was, so I'd change it to this:

```python
"pattern: \"{pattern}\" could not be compiled as a \"{retype}\" instance"
```
```python
"error compiling pattern '{pattern}': {e}".format(pattern=pattern, e=format_exception(e))
```

aodncore/util/misc.py
```diff
@@ -95,6 +96,24 @@ def format_exception(exception):
...
+                pattern=pattern,
+                retype=retype))
+
+    file_basename = os.path.basename(filepath)
```
lwgordonimos: This is definitely a potential point of confusion... if you have a path and a regex it's not intuitive that inside the function, it would silently match against the basename.

Maybe we should add a 'basename' parameter which defaults to False, so that if you want to match against only the basename, it's obvious to the caller.

test_aodncore/util/test_misc.py
```diff
@@ -211,6 +212,34 @@ def test_validate_relative_path_attr(self):
...
+        # test on filepath
+        fields = get_fields_from_filename(os.path.join(TEST_ROOT, good_file),
+                                          good_pattern)
+        self.assertEqual(fields['product_code'], 'ABC')
+
+        with self.assertRaises(TypeError):
+            get_fields_from_filename(bad_file, 12)
+
+        fields = get_fields_from_filename(bad_file, '')
+        self.assertTrue(not fields)
```
lwgordonimos: I'd update this test to:
```python
self.assertDictEqual(fields, {})
```
to be more specific

test_aodncore/util/test_misc.py
```diff
@@ -211,6 +212,34 @@ def test_validate_relative_path_attr(self):
...
+                                       """, re.VERBOSE)
+        good_file = 'FILE_SUFFIX_ABC_C-20180101T000000Z.txt'
+        bad_file = 'BAD_ABC_C-20180101T000000Z.txt'
+
+        # test on filename only
+        fields = get_fields_from_filename(good_file, good_pattern)
+        self.assertEqual(fields['product_code'], 'ABC')
+
+        # test on filepath
+        fields = get_fields_from_filename(os.path.join(TEST_ROOT, good_file),
```
lwgordonimos: I wouldn't worry about joining this to a real path because that kind of implies the file actually exists. I'd just use a dummy "/not/a/real/path" kind of thing so it's obvious that only the path matters and not an actual file.

aodncore/util/misc.py
lbesnard: I guess this is why the name of the function is
get_fields_from_filename

Doing the os.path.basename was to force this in case a full path is the input of the function

aodncore/util/misc.py
```diff
@@ -95,6 +96,22 @@ def format_exception(exception):
     return "{cls}: {message}".format(cls=exception.__class__.__name__, message=exception)
 
 
+def get_fields_from_str(str, pattern):
```
lwgordonimos: str is a reserved word, so needs to be changed. something like "input_string". Apart from that, I can't think of a better more concise name, so happy to merge!

lwgordonimos: get_pattern_subgroups_from_string?

aodncore/util/misc.py
```diff
@@ -95,6 +96,22 @@ def format_exception(exception):
     return "{cls}: {message}".format(cls=exception.__class__.__name__, message=exception)
 
 
+def get_pattern_subgroups_from_string(str, pattern):
```
mhidas: Why _sub_groups? I think `get_pattern_groups_from_string` would be fine, or even `get_pattern_dict_from_string`.

lbesnard: we just changed that with Leigh, but we can have an un-official meeting with @lwgordonimos ?

aodncore/util/misc.py
```diff
@@ -95,6 +96,22 @@ def format_exception(exception):
...
+    """Function to retrieve parts of a string given a compiled pattern (re.compile(str))
```
mhidas: Don't you mean (re.compile(pattern)) here?

aodncore/util/misc.py
lwgordonimos: It's the terminology used by the re library:
https://docs.python.org/2/library/re.html#re.MatchObject.groupdict

The primary use case is "fields from an IMOS filename", but this is really a generic function to get any "named subgroups" from a "string", so I was thinking to not imply that it is coupled tightly to that use case.

If there's a need for a wrapper specifically for *fields* it could also do validation against http://content.aodn.org.au/Documents/IMOS/Conventions/IMOS_NetCDF_Conventions.pdf it could be named in line with those conventions?

I'm not overly precious about it, but that was the thinking anyway!

aodncore/util/misc.py
```diff
@@ -95,6 +96,22 @@ def format_exception(exception):
...
+    """
+    retype = type(re.compile(''))
+    if not isinstance(pattern, retype):
+        try:
+            pattern = re.compile(pattern)
+        except Exception as err:
+            raise TypeError("error compiling pattern '{pattern}': {e}".
+                            format(pattern=pattern, e=format_exception(err)))
+
+    m = pattern.match(string)
```
mhidas: Note that this will only return anything if the pattern matches the _beginning_ of the string (see https://docs.python.org/2/library/re.html#re.RegexObject.match). So
* there's no need to start the pattern with "^"; and
* if you want to match _anywhere_ in the string, start the pattern with ".*".

Maybe worth noting this in the doc string?

lbesnard:
@mhidas done

mhidas:
This has been included in 0.13.0 and tested in RC. Seems to work ok.

PR #104, opened 2018-06-18
Add optional configuration item 'opendap_root' to this to be set on a…

lwgordonimos:
… *per-environment* basis


lbesnard:
re the value to add on chef-private
```http://thredds.aodn.org.au/thredds/dodsC/```

PR #105, opened 2018-06-21
Fix behaviour of input file archiving so that simply accessing the fi…

lwgordonimos:
…le object doesn't trigger archiving

https://github.com/aodn/python-aodndata/pull/55 made me realise the core behaviour was not correct for the input_file_object I suggested to use, so fixing here!

### DIFF
--- a/aodncore/pipeline/handlerbase.py
+++ b/aodncore/pipeline/handlerbase.py
@@ -653,7 +653,8 @@ class HandlerBase(object):
             self._upload_runner_archive.run(files_to_archive)
 
         if self.archive_input_file:
-            self.input_file_object.publish_type = PipelineFilePublishType.ARCHIVE_ONLY
+            if self.input_file_object.publish_type is PipelineFilePublishType.UNSET:
+                self.input_file_object.publish_type = PipelineFilePublishType.ARCHIVE_ONLY
             self.input_file_object.archive_path = self.input_file_archive_path
             self._upload_runner_archive.run(self.input_file_object)
 
--- a/aodncore/pipeline/handlerbase.py
+++ b/aodncore/pipeline/handlerbase.py
@@ -334,8 +334,8 @@ class HandlerBase(object):
         self._notification_results = None
         self._should_notify = None
         self._state_query = None
-        self._upload_runner_object = None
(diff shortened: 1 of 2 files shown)


aodncore/pipeline/handlerbase.py
```diff
@@ -641,6 +639,7 @@ def _archive(self):
             self._upload_runner_archive.run(files_to_archive)
 
         if self.archive_input_file:
+            self.input_file_object.publish_type = PipelineFilePublishType.ARCHIVE_ONLY
```
mhidas: I think this is ok, except in the (unlikely?) case I want to e.g. archive _and harvest_ the input file. I can still do that, just not using the `archive_input_file`, because if I set that this will then clobber the actual publish_type I want (set elsewhere).

I guess we could add something like 
```python
if self.input_file_object.publish_type is PipelineFilePublishType.UNSET
```

lwgordonimos:
@mhidas I added the non-clobbering if statement, because I think that's sensible.

This has however required a change to the logger/upload runner attributes to make testing easier, since setting the attribute caused a "chicken and egg" scenario with calling the logger. Making them lazy properties instead using the initialise step resolves this and is a better pattern anyway.

aodncore/pipeline/handlerbase.py
```diff
@@ -609,20 +622,51 @@ def temp_dir(self):
...
+        """
+        if self._upload_runner_object is None:
+            self._upload_runner_object = get_store_runner(self._config.pipeline_config['global']['upload_uri'],
+                                                          self._config, self.logger)
+            self.logger.sysinfo("get_store_runner (upload)-> {runner}(broker='{broker}')".format(
+                runner=self._upload_runner_object.__class__.__name__, broker=self._upload_runner_object.broker.__class__.__name__))
+        return self._upload_runner_object
+
+    @property
+    def _upload_runner_archive(self):
```
mhidas: For simplicity and consistency, could this just be called `_archive_runner`?

lwgordonimos: Yes good point, this is a relic of before this was refactored into the current store runner/broker design.

I've updated to reflect what they actually represent now!

PR #112, opened 2018-08-07
Get rid of pip dependency from setup.py

lwgordonimos:


lwgordonimos:
This was originally due to the pip 10 breaking the "pip.req" usage, which was because you're actually not supposed to use that API, as it's not public, then I found that article which makes reasonable sense, and I didn't want to just duplicate the requirements in two places.

It's still not overly intuitive, but it at least puts the emphasis on packaging by consolidating around setup.py, and requirements.txt is kind of a bonus for local virtualenvs etc.

PR #121, opened 2018-09-10
Fix delete_regexes method to compare pattern strings, due to unreliab…

lwgordonimos:
…le comparison of Pattern types


mhidas:
:+1: Third time lucky!
Thanks

PR #126, opened 2018-10-02
Remove cc-plugin-imos dependency

lwgordonimos:
Goes with: https://github.com/aodn/python-aodndata/pull/88

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -4,8 +4,7 @@ from aodncore.version import __version__
 
 ENTRY_POINTS = {
     'pipeline.module_versions': [
-        'compliance-checker = compliance_checker:__version__',
-        'cc-plugin-imos = cc_plugin_imos:__version__'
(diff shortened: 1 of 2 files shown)


mhidas:
It might be an issue with my PyCharm setup, but after switching to this branch and uninstalling cc-plugin-imos from my virtualenv, I get 34 failing tests in `test_dummyHandler`, which seem to be related to the plugin...
`DistributionNotFound: The 'cc-plugin-imos>=1.3.0' distribution was not found and is required by the application`

mhidas:
Never mind. If I run it in the shell it works fine. I'll have to update something in PyCharm.

mhidas:
@lwgordonimos I've built this (as v0.21.1) and moved it to systest, so we'll see how it goes goes getting deployed to RC.

PR #129, opened 2018-10-15
Update paramiko due to CVE-2018-1000805

lwgordonimos:


mhidas:
I assume this is ready to merge?

lwgordonimos:
Should be. Nothing actually uses SFTP anywhere yet, and this is a minor release to fix that CVE (which GitHub complained about!).

PR #137, opened 2018-11-20
Filetype multiple extensions

lwgordonimos:
This allows multiple extensions to correspond with a single FileType enum member.

The primary change is that the "extension" attribute of the enum members has changed from a string to a tuple of valid extensions for a given type. Additionally, the comparison is now case-insensitive, which means for example jpg, jpeg, JPG and JPEG will all be interpreted as a `FileType.JPEG`.

Since this is no longer a one-to-one mapping between extension and type, it's no longer possible to return a single extension from a type, but a) I couldn't find any code doing this and b) if this is required for some reason, it could be a convention to use the *first* extension as a default or preferred option. Hard to envisage anything wanting to do this though, but is hypothetically possible.

I couldn't see anything in python-aodndata that should be affected by this, since comparisons such as `PipelineFile.file_type is FileType.JPEG` are unaffected, and that is the predominant use of this enum.

### DIFF
--- a/aodncore/pipeline/steps/resolve.py
+++ b/aodncore/pipeline/steps/resolve.py
@@ -50,19 +50,19 @@ def get_resolve_runner(input_file, output_dir, config, logger, resolve_params=No
     :param resolve_params: dict of parameters to pass to :py:class:`BaseResolveRunner` class for runtime configuration
     :return: :py:class:`BaseResolveRunner` class
     """
-    _, file_extension = os.path.splitext(input_file)
+    file_type = FileType.get_type_from_name(input_file)
 
-    if file_extension == FileType.ZIP.extension:
+    if file_type is FileType.ZIP:
         return ZipFileResolveRunner(input_file, output_dir, config, logger)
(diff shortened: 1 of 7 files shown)


aodncore/pipeline/steps/resolve.py
```diff
@@ -52,17 +52,17 @@ def get_resolve_runner(input_file, output_dir, config, logger, resolve_params=No
     """
     _, file_extension = os.path.splitext(input_file)
```
mhidas: Why not just `file_type = FileType.get_type_from_name(input_file)`?

lwgordonimos: Yep, definitely should, changed. 

I think the extension based thing was from before the FileType enum evolved into it's current form, but now it's just redundant.

mhidas: Cool, though I thought you could still store the result in a variable to avoid calling `FileType.get_type_from_name(input_file)` repeatedly...

PR #139, opened 2018-11-22
make typename a positional arg of WfsBroker.getfeature_dict()

mhidas:
SUGGESTION ONLY... may need more work  before merge.

### DIFF
--- a/aodncore/util/wfs.py
+++ b/aodncore/util/wfs.py
@@ -54,14 +54,16 @@ class WfsBroker(object):
         """
         return self._wfs
 
-    def getfeature_dict(self, **kwargs):
+    def getfeature_dict(self, typename, **kwargs):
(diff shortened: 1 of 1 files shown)


lwgordonimos:
I think this kind of constraint should be added to a separate higher level method. 

This should be as generic as possible with the only constraint being the JSON output to map into a dict. Then if you want helper methods with *more* constraints (e.g. making the layer a required parameter) they can be a *wrapper* for this low level request.

This should probably then be a "private" method, and the public helper method(s) can then add constraints/specific behaviour on top of that. Which achieves your goal in a slightly different way and keeps the internals library code more flexible and generic.

mhidas:
This hasn't actually been an issue, and nothing in aodndata is calling this `getfeature_dict` method directly, so we can just drop it.

PR #140, opened 2018-11-28
Add TIFF file type and is_image_type property

mhidas:


mhidas:
This would be useful for https://github.com/aodn/python-aodndata/pull/93

aodncore/pipeline/common.py
```diff
@@ -155,6 +156,11 @@ def get_type_from_name(cls, name):
         _, extension = os.path.splitext(name)
         return cls.get_type_from_extension(extension)
 
+    @property
+    def is_image(self):
```
lwgordonimos: I think the concept is good. The only thing I would suggest changing here would be to rename this to to more explicit that it is asking a question about the current *type* , i.e. `is_image_type`.

`is_image` as a name overlaps with the functions that actually *do* check the contents of the file, so this could be ambiguous in terms of what is being asked.

```python
my_file_type = FileType.JPEG
is_this_an_image_type = my_file_type.is_image_type
```

mhidas: :+1: Done.

lwgordonimos:
@mhidas very slight refactor to make this feature more generic, but I thought image in particular was a common enough check to keep a helper method just for that which wraps the underlying check

test_aodncore/pipeline/test_common.py
```diff
@@ -38,3 +38,10 @@ def test_get_type_from_name_unknown(self):
         random_filename = "file.{}".format(str(uuid.uuid4()))
         unknown_type = FileType.get_type_from_name(random_filename)
         self.assertIs(unknown_type, FileType.UNKNOWN)
+
+    def test_is_image(self):
```
mhidas: Forgot to rename this to `test_is_image_type`... will fix.

PR #145, opened 2018-12-11
Modernize, add Python 3 support

lwgordonimos:


aodncore/util/fileops.py
```diff
@@ -137,7 +138,7 @@ def is_dir_writable(path):
     """
     try:
         with TemporaryFile(prefix='is_dir_writable', suffix='.tmp', dir=path) as t:
-            t.write('is_dir_writable')
+            t.write('is_dir_writable'.encode('utf-8'))
```
mhidas: Is `'x'.encode('utf-8')` the same as `u'x'`?

mhidas:
Is this ready to merge?

aodncore/pipeline/handlerbase.py
```diff
@@ -527,13 +528,14 @@ def logger(self):
         return self._logger
 
     @lazyproperty
-    def module_versions(self):
+    def versions(self):
```
mhidas: This isn't called explicitly anywhere?

aodncore/vocab/platform_code_vocab.py
```diff
@@ -80,7 +80,7 @@ def platform_type_uris_by_category(self):
                     # handle more than 1 url match per category of platform
                     if platform_element_sublabels is not None:
                         if 'narrowMatch' in platform_element_sublabels:
-                            val_cat_url = val.attrib.values()[0]
+                            val_cat_url = list(val.attrib.values())[0]
```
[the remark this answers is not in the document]
lwgordonimos: Yeah any number of these could be written differently, but for example, that would result in changing the exception type from IndexError to StopIteration, so the minimal "compatible" change was my priority here, without fixing or optimising anything at the same time.

aodncore/util/fileops.py
lwgordonimos: It actually needs a "bytes-like" object not a unicode. Could have been written as a b-string, but results in the same thing.

```python
In[20]: b'is_dir_writable' == 'is_dir_writable'.encode('utf-8')
Out[20]: True
```

aodncore/pipeline/handlerbase.py
lwgordonimos: Nope, is only used when representing the handler as a string, i.e. in the logs

lwgordonimos:
Yeah is mergable as far as I have tested, haven't found any issues so far spinning up on Python 2 or 3 in a test environment.

aodncore/util/fileops.py
mhidas: Ah! Right, thanks. Still getting my head around bytes/string/unicode objects.

PR #146, opened 2018-12-13
Add a log message to make it clearer when users are excluded from not…

lwgordonimos:
…ification

This makes it clearer as to why the configured error recipients are *not* notified when the error type is not a PipelineProcessingError (which is the only type of exception which is forwarded to end users)


bpasquer:
thanks @lwgordonimos 👍

lwgordonimos:
Yeah if you're happy with the wording, since it is intended to be informative to POs!

mhidas:
Ah, right. In that case something like "exception is not a user-correctable problem, ..." might be more informative?

lbesnard:
Maybe we should just merge this !

lwgordonimos:
@lbesnard @mhidas yeah, I just rebased and updated to Marty's wording, so should be right to merge now

PR #148, opened 2019-02-13
Fix bug causing handler configuration warnings to be hidden

lwgordonimos:
When accessing dict items using str.format, the key should *not* be quoted.

This caused legitimate configuration warnings to be hidden by a useless KeyError.


lwgordonimos:
@bpasquer this PR solves the problem with hiding useful errors.

PR #152, opened 2019-04-03
DONT_MERGE: example implementation of a non-Talend "harvester"

ghost:


mhidas:
This is no longer needed, as we now have the CSV harvester which is actually a functional harvester written in Python.

PR #154, opened 2019-04-24
Jenkinsfile

pmbohm:
aodn/backlog#1118


requirements.txt
```diff
@@ -1,2 +1,3 @@
 --index-url https://pypi.python.org/simple/
--e .
\ No newline at end of file
+-e .
+cc-plugin-imos==1.3.6
```
mhidas: This should not be needed for aodncore. Nothing in here refers to the IMOS checks.

PR #157, opened 2019-06-20
Update platform vocab code to source URLs from config file

lwgordonimos:
There is a current issue where testing this code requires either mocking out the loading of the vocabs or polluting the handlers with extra code.

https://github.com/aodn/python-aodndata/blob/master/test_aodndata/soop/test_soop_tmv_nrt.py#L128
https://github.com/aodn/python-aodndata/blob/master/aodndata/soop/soop_co2.py#L65


lbesnard:
@lwgordonimos I'm merging but not sure about this drastic decrease of code coverage, just saying

PR #158, opened 2019-07-01
Implement custom JSON encoding/decoding to (de)serialise Enum, Pipeli…

lwgordonimos:
…neFile and PipelineFileCollection objects

### DIFF
--- a/aodncore/pipeline/common.py
+++ b/aodncore/pipeline/common.py
@@ -1,7 +1,6 @@
 """This module contains common "types" (typically enums or simple data classes) used by more than one module in the
     project.
 """
-
 import os
 
 from enum import Enum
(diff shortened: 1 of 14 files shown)


lwgordonimos:
This was a proof of concept to chain handlers together (for products I think), but is likely a dead-end. It is bad for scalability, as it makes assumptions about the local filesystem and the initial problem was solved differently, so this should probably just be closed. Has some interesting examples/concepts in it, re: using exit policies for more things, but I wouldn't merge it.

mhidas:
> This was a proof of concept to chain handlers together (for products I think), but is likely a dead-end. It is bad for scalability, as it makes assumptions about the local filesystem and the initial problem was solved differently, so this should probably just be closed. Has some interesting examples/concepts in it, re: using exit policies for more things, but I wouldn't merge it.

I agree. Let's close it, but keep the branch for future reference.

PR #163, opened 2019-07-30
Remove numpy dependency, as this package does not directly depend on it

lwgordonimos:


lwgordonimos:
@mhidas Yes, should be good to go. It was initially going to pin the numpy version to <1.17.0 because that breaks Python 2, but then I realised it actually isn't a dependency at all.

PR #164, opened 2019-08-05
read event ids for incoming files from uuid_path if it exists

craigrose:
This adds the feature of being able to provide event-ids from an external source.  This allows for tracking of progress of a file through the pipeline externally via the logs.


lwgordonimos:
Closing as it was identified that library changes are undesirable to support testing, and the following PR was updated accordingly: https://github.com/aodn/utilities/pull/166

PR #167, opened 2019-08-28
Add file download feature for product handlers

lwgordonimos:


aodncore/pipeline/files.py
```diff
@@ -478,31 +604,31 @@ def add(self, pipeline_file, deletion=False, overwrite=False):
...
         validate_bool(overwrite)
 
-        if isinstance(pipeline_file, PipelineFile):
+        if isinstance(pipeline_file, self.member_class):
             fileobj = pipeline_file
         else:
             if not deletion and not os.path.isfile(pipeline_file):
                 raise MissingFileError("file '{src}' doesn't exist".format(src=pipeline_file))
-            fileobj = PipelineFile(pipeline_file, is_deletion=deletion)
+            fileobj = self.member_class(pipeline_file, is_deletion=deletion)
```
mhidas: I get an error here when trying to create a `RemotePipelineFileCollection`. 
```
  File "/sw/chef/src/python-aodndata/aodndata/moorings/products_handler.py", line 88, in preprocess
    input_file_collection = RemotePipelineFileCollection(input_files)
  File "/sw/chef/src/python-aodncore/aodncore/pipeline/files.py", line 950, in __init__
    super(RemotePipelineFileCollection, self).__init__(*args, **kwargs)
  File "/sw/chef/src/python-aodncore/aodncore/pipeline/files.py", line 578, in __init__
    self.add(f)
  File "/sw/chef/src/python-aodncore/aodncore/pipeline/files.py", line 616, in add
    fileobj = self.member_class(pipeline_file, is_deletion=deletion)
TypeError: __init__() got an unexpected keyword argument 'is_deletion'
```
Interestingly, `is_deletion` is technically _only_ relevant to a remote file, not a local one, but that's not the way the handler deals with deletions...

aodncore/pipeline/storage.py
```diff
@@ -101,6 +112,51 @@ def set_is_overwrite(self, pipeline_files, dest_path_attr='dest_path'):
...
+        :return: None
+        """
+        download_collection = ensure_remotepipelinefilecollection(remote_pipeline_files)
+
+        self._pre_run_hook()
+
+        for remote_pipeline_file in download_collection:
+            try:
+                self._download_file(remote_pipeline_file=remote_pipeline_file, local_path=local_path)
+                remote_pipeline_file.local_path = os.path.join(local_path, self.prefix, remote_pipeline_file.dest_path)
```
mhidas: I don't think this line is quite right: the `self.prefix` trumps the `local_path` before it.
Actually `LocalFileStorageBroker._download_file()` already sets the `local_path` attribute, so at least for that case, you don't need to do it here too.
You also don't have this line in `download_iterator()` below (line 150).

PR #171, opened 2019-10-17
Update discover_entry_points function to return a list of failed entr…

lwgordonimos:
…y points instead of raising Exception

This prevents the scenario where the handler will raise an unhandled exception if one of the dependencies has a problem with one of the entry points defined by an external package (e.g. aodndata, aodntools, cc_plugin_imos).

### DIFF
--- a/aodncore/pipeline/destpath.py
+++ b/aodncore/pipeline/destpath.py
@@ -86,13 +86,14 @@ class PathFunctionResolver(object):
     def _resolve_from_param(self):
         if isinstance(self.function_param, six.string_types):
             # a string parameter is assumed to be referring to an advertised entry point of the same name
+            loaded_functions, failed_functions = self.handler_instance.config.discovered_dest_path_functions
             try:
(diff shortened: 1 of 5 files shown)


aodncore/pipeline/destpath.py
```diff
@@ -86,13 +86,14 @@ def _resolve_from_handler(self):
     def _resolve_from_param(self):
         if isinstance(self.function_param, six.string_types):
             # a string parameter is assumed to be referring to an advertised entry point of the same name
+            loaded_functions, failed_functions = self.handler_instance.config.discovered_dest_path_functions
```
mhidas: Should there be at least a warning if there are any `failed_functions`?

lwgordonimos: This class is only concerned with a single *current* function, so doesn't make sense to warn about unrelated functions. I have however updated the exception message to differentiate between missing and failed functions

mhidas: :+1:

PR #186, opened 2019-12-02
Pin python-dateutil

lwgordonimos:


lwgordonimos:
Closing, as we're going to handle this with an environment level requirements.txt rather than pinning things here.

PR #187, opened 2019-12-03
Update worker logging code to only supply each task with it's *own* l…

lwgordonimos:
…ogging config

### DIFF
--- a/aodncore/pipeline/configlib.py
+++ b/aodncore/pipeline/configlib.py
@@ -25,9 +25,9 @@ __all__ = [
     'validate_lazyconfigmanager'
 ]
 
-DEFAULT_CONFIG_FILE = '/mnt/ebs/aodn-pipeline/etc/pipeline.conf'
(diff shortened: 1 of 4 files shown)


lwgordonimos:
@mhidas this looks to work well on my dev environment. 

1. Log files are opened lazily, when the logger receives a message
1. Each worker only opens it's own log file

```
$ sudo lsof -u pipeline_worker | grep tasks
celery  17479 pipeline_worker    6w      REG  252,0     5176 1977078 /mnt/ebs/log/pipeline/process/tasks.ANMN_AM.log
celery  17662 pipeline_worker    6w      REG  252,0    10507 1977077 /mnt/ebs/log/pipeline/process/tasks.SRS_OC_BODBAW.log
```

mhidas:
I've built and promoted this to RC, so we can test it. So because there is a code change, all the pipelines will be restarted, right?

lwgordonimos:
Yes that's right.

PR #188, opened 2019-12-04
Add additional "late deletion" step to TalendHarvestRunner to allow d…

lwgordonimos:
…eletions to occur *after* additions

### DIFF
--- a/aodncore/pipeline/files.py
+++ b/aodncore/pipeline/files.py
@@ -205,15 +205,17 @@ class PipelineFile(PipelineFileBase):
     :type dest_path: :py:class:`str`
     :param is_deletion: flag designating whether this is a deletion
     :type is_deletion: :py:class:`bool`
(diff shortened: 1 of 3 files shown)


lwgordonimos:
@mhidas this works pretty well in unit tests, so I think ready for review.

Under this model, publishing with the Talend runner is now performed in 3 phases:

1. early deletions
1. additions
1. late deletions

Note that non-Talend UPLOAD_ONLY/DELETE_ONLY operations are still run after these steps (unchanged).

```python
pf1 = PipelineFile('/path/to/file1.nc', is_deletion=True)
pf2 = PipelineFile('/path/to/file2.nc', is_deletion=True, late_deletion=True)
```

The basic usage is as above, which would cause pf1 will be "unpublished" before additions are run, and pf2 will not be unpublished until after additions are *successfully* run.

Note that if the additions fail, the late deletions will not be run, so will leave the late deletions collection published in the event of an error.

PR #192, opened 2019-12-10
unpin version of owslib in setup.py

mhidas:
More unpinning required: OWSLib is currently inconsistent with the [pipeline_requirements.txt](https://github.com/aodn/chef/blob/master/cookbooks/imos_po/files/default/pipeline_requirements.txt#L6) file in chef and therefore breaking all the pipelines :(


mhidas:
Good to go!

mhidas:
Thanks! I'll build it.

PR #199, opened 2019-12-16
Make WFS version configurable

lwgordonimos:


aodncore/util/wfs.py
```diff
@@ -8,11 +8,14 @@
 from ..util import IndexedSet
 
 __all__ = [
+    'DEFAULT_WFS_VERSION',
     'WfsBroker',
     'get_filter_for_file_url',
     'ogc_filter_to_string'
 ]
 
+DEFAULT_WFS_VERSION = '1.0.0'
```
mhidas: Shouldn't this be "1.1.0"?

test_aodncore/pipeline/test_configlib.py
```diff
@@ -19,7 +19,8 @@
         'opendap_root': 'http://opendap.example.com',
         'processing_dir': '/tmp/probably/doesnt/exist/processing',
         'upload_uri': 'file:///tmp/probably/doesnt/exist/upload',
-        "wfs_url": "http://geoserver.example.com/geoserver/wfs",
+        'wfs_url': 'http://geoserver.example.com/geoserver/wfs',
+        'wfs_version': '1.0.0',
```
mhidas: And here too? 1.1.0?

lwgordonimos:
@mhidas this is the first step in updating to WFS 1.1.0. Note that the default is still 1.0.0 here, since updating will break aodndata.

The underlying library calls are quite different, so mocking with httpretty isn't really feasible (since it's pretty good at mocking a predictable sequence of HTTP requests, but pretty bad if the requests vary at all, which they do when changing WFS versions). For this reason it mocks with `mock` again, which avoids all that low level HTTP stuff.

aodncore/util/wfs.py
lwgordonimos: See my comment below. If we immediately updated it to 1.1.0, it would break aodndata until https://github.com/aodn/python-aodndata/pull/197 is merged/built/deployed.

mhidas:
> @mhidas this is the first step in updating to WFS 1.1.0. Note that the default is still 1.0.0 here, since updating will break aodndata.

Ah, ok!

PR #202, opened 2020-01-13
Refactor Celery specific references out of HandlerBase into the calli…

lwgordonimos:
…ng PipelineTask class to support non-Celery use cases


mhidas:
@lwgordonimos I guess this one's just an idea for future development?

lwgordonimos:
Yes that's right. It's very incomplete, and well out of date, but was just playing with getting celery references out of the core code. Not sure if there's anything really useful or salvageable from it to be honest...

PR #206, opened 2020-02-26
Switch to WFS version 1.1.0

mhidas:
Now that https://github.com/aodn/python-aodndata/pull/197 has been merged and promoted to prod, I think we're ready for this!


lwgordonimos:
@mhidas looks good, should be right to test in RC.

Looks like the only two handlers using this so far are these two, so should be straight-forward to verify:

SoopTrvHandler
MooringsProductsHandler

PR #211, opened 2020-05-19
WfsBroker methods to accept OgcExpression filters not XML

mhidas:
When querying WFS, we construct filters using `owslib.fes.OgcExpression` objects. We only need to convert them to XML when we send the getFeature request. So I think this should just be done "under the hood" automatically, rather than requiring the hanlder to do it (e.g. [here](https://github.com/aodn/python-aodndata/blob/8de9b2bb20502ccea8407cd8f59ed87525969ce4/aodndata/moorings/products_handler.py#L166)).
Also, I think the use of `ogc_filter` to denote both `OgcExpression` objects and XML was confusing.

### DIFF
--- a/aodncore/pipeline/files.py
+++ b/aodncore/pipeline/files.py
@@ -208,29 +208,31 @@ class PipelineFile(PipelineFileBase):
     :type late_deletion: :py:class:`bool`
     :param file_update_callback: optional callback to call when a file property is updated
     :type file_update_callback: :py:class:`callable`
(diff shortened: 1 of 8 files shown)


mhidas:
@lwgordonimos What do you think? Could we add something along these lines to your refactoring PR?

lwgordonimos:
Good ideas. I've added it to the "general client API improvements PR": https://github.com/aodn/python-aodncore/pull/209

PR #212, opened 2020-07-30
Add validate_unique parameter to PipelineFileCollection, to allow dup…

lwgordonimos:
…licate checking to be disabled

### DIFF
--- a/aodncore/pipeline/files.py
+++ b/aodncore/pipeline/files.py
@@ -569,12 +569,13 @@ class PipelineFileCollectionBase(MutableSet):
 
     :param data: data to add during initialisation of the collection, either a single :py:class:`PipelineFile` or file
         path, or an :py:class:`Iterable` whose elements are :py:class:`PipelineFile` instances or file paths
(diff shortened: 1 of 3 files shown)


kereid:
resolved by https://github.com/aodn/issues/issues/851

anguss00 (approved):

mhidas:
@lwgordonimos @lbesnard Was this PR actually needed for anything? The updated version of aodncore (1.1.5) was never even deployed to RC, let alone prod!

Just asking because I've just merged #216, which will improve legibility of the pipeline logs, and it would be nice to test & deploy that (now v1.1.6).

mhidas:
For now I've just promoted 1.1.5 to RC

lwgordonimos:
@mhidas it is just a performance optimisation, so should be transparent (other than being faster for large collection sizes). So other than general regression testing, it shouldn't need any special attention.

mhidas:
:+1: Ok. I've tested and promoted 1.1.5 to prod.

PR #215, opened 2020-09-04
Add support for CSV delete manifests

lwgordonimos:


aodncore/pipeline/steps/resolve.py
```diff
@@ -193,18 +206,25 @@ class MapManifestResolveRunner(BaseManifestResolveRunner):
...
-            for line_newline in f:
-                line = line_newline.rstrip(os.linesep)
-                src, dest_path = line.split(self.PATH_SPLIT_CHAR, 1)
-
-                abs_path = self.get_abs_path(src)
-                fileobj = PipelineFile(abs_path, dest_path=dest_path)
-
-                self._collection.add(fileobj)
+        table = tableschema.Table(self.input_file, headers=self.schema.field_names, schema=self.schema, format='csv')
+        for local_path, dest_path in table.iter(exc_handler=self._exc_handler):
```
mhidas, line 224: Don't we need to handle schema errors here too, as you've done in `DeleteManifestResolveRunner`? 
e.g. if either the local_path or dest_path are empty?
Or are these cases adequately handled by `PipelineFile.__init__`?

aodncore/pipeline/steps/resolve.py
```diff
@@ -318,6 +338,52 @@ def run(self):
...
 
 
+class DeleteManifestResolveRunner(BaseManifestResolveRunner):
+    """Handles a delete manifest file which only contains a list of source files, and a valid "deletion publish type"
+        string, as defined by the PipelineFilePublishType enum
+
+    File format must be as follows::
+
+        destination/path/for/delete1,DELETE_PUBLISH_TYPE
+        destination/path/for/delete2,DELETE_PUBLISH_TYPE
```
mhidas, line 348: I hadn't thought of specifying the publish type in the manifest. I guess this could be useful, but I think it most cases it would be up to the specific handler to determine what it needs to do. Could we allow this to be left blank, in which case the handler decides?

mhidas (commented):
:+1:  This is neat. I just have one main query.

Good to see `tableschema` make a first appearance!

aodncore/pipeline/steps/resolve.py
lwgordonimos: Yeah good point, I think an empty string would end up with an error which wouldn't be handled.

aodncore/pipeline/steps/resolve.py
lwgordonimos: Yeah that's a fair point. I was thinking in terms of the *person* deciding when they create the manifest but you make a good point about the handler logic being able to decide on a file by file basis (as it often does for additions!).

aodncore/pipeline/steps/resolve.py
lwgordonimos: @mhidas I've refactored the validation logic into a base class which I think should make it flexible enough to handle pretty much any kind of CSV we want to handle. The child classes define the schema of the CSV, and a "row handler" method which translates each row into a PipelineFile.

aodncore/pipeline/steps/resolve.py
lwgordonimos: @mhidas I've addressed this by removing the `required` constraint and not setting the `publish_type` property unless it has been provided for any given row.

https://github.com/aodn/python-aodncore/pull/215/files#diff-dc09111dc00c10cc36c7d376aa0a09095f8aa1503b66cba58d7d31c4e4e79820R2

The only slightly quirky thing with having this optional is that a trailing comma is **required** if the publish type field is omitted, otherwise the row is parsed as only having 1 field instead of 2. This gives the most control/flexibility to people in setting the publish type, with the only caveat being that minimal delete manifests look like this:

```csv
PATH/TO/FILE/1,
PATH/TO/FILE/2,
PATH/TO/FILE/3,
```

So we really have two options which I'm happy to go with either, depending on the usability from your side:

- Option A: if the commas are annoying, remove the publish_type field from the schema and have a single field "CSV" (but the handler will *always* be solely responsible for handling the publish type)
- Option B: leave this feature, and POs creating manifests need to remember the trailing comma (if omitted, the file will land in error and the schema error will be in the process log)

aodncore/pipeline/common.py
```diff
@@ -139,9 +139,10 @@ class Object(object):
     NETCDF = (('.nc',), 'application/octet-stream', is_netcdf_file)
     DIR_MANIFEST = (('.dir_manifest',), 'text/plain', is_nonempty_file)
     JSON_MANIFEST = (('.json_manifest',), 'application/json', is_json_file)
-    MAP_MANIFEST = (('.map_manifest',), 'text/plain', is_nonempty_file)
+    MAP_MANIFEST = (('.map_manifest',), 'text/csv', is_nonempty_file)
     RSYNC_MANIFEST = (('.rsync_manifest',), 'text/plain', is_nonempty_file)
     SIMPLE_MANIFEST = (('.manifest',), 'text/plain', is_nonempty_file)
+    DELETE_MANIFEST = (('.delete_manifest',), 'text/csv', is_nonempty_file)
```
mhidas, line 145: This just reminded me that we're actually doing validation here. Would it make sense to do the `tableschema` validation here, rather than as part of the resolve step?

Just a thought. From a pipeline's perspective I don't think it makes a difference either way, so probably not worth the refactoring.

aodncore/pipeline/steps/resolve.py
mhidas: @aodn/data-team Any thoughts?

Mine are:
* So far the old `po_s3_del` command has satisfied all my deletion needs (equivalent to `DELETE_UNHARVEST` publish_type, and silently ignoring cases where there's no harvester to unharvest a file), so Option A would pr

aodncore/pipeline/common.py
lwgordonimos: In terms of this Enum, all it should only be doing is very lightweight file *type* validation and not *content*, so shouldn't be reading the full file contents. It would also make it start to cross boundaries into the workflow side of things, when it's supposed to just be a generic utility enum and deliberately kept lightweight and *not* reading the full file content.

To that end, I think the fact that I added the `is_json_file` was incorrect and should be removed in favour of `is_nonempty_file`, since that loads the whole file. 

Also, and probably most importantly, the way the tableschema validation is written (in line with upstream tableschema doco), is that uses an iterator for **streaming** the CSV input file, and validating on a row-by-row basis (which is also good for scalability reasons), so validating the file independently from "resolving" it also doesn't fit the "validate on the fly" model which tableschema implements. It wouldn't make sense to loop through the whole (possibly large) file twice, once to validate it, and the second time to load the manifest into a collection.

So given all that, the original intent, and IMO still the optimal approach for now is:

* `FileType` enum = cheap type validation (e.g. read file header *only*, to pick up obvious errors like a `.png` actually being a renamed `.jpg` for example) - basically the equivalent of the Linux `file` command
* `*ResolveRunner` = full content validation, optimised for scalability

aodncore/pipeline/steps/resolve.py
lwgordonimos: YAGNI principle would dictate that we retain the current behaviour unless the feature is demonstrated to be useful to anyone. Would be preferable to adding complexity for a feature nobody uses!

aodncore/pipeline/common.py
mhidas: :+1: Thanks, that all makes sense, and I agree.

There is also `*CheckRunner`, which for now is only used to do (some) non-trivial _content_ validation for NetCDF files. In the future there could be additional check runners defined to check that e.g. a PDF document, or PNG image is not corrupted.

lwgordonimos: Yes, for completeness the list of various things performing file validation as it stands and the distinction between them is as follows (after a quick poke around to refresh my memory!):

* `*ResolveRunner` = validation and resolution of the **input file** (optionally full content validation, depending on implementation, e.g. manifest files)

* `*CheckRunner` =  validation of all **resolved files in the handler collection** which define a `check_type` (optionally full content validation, depending on check type)

* `FileType` = basic file format validation for **resolved files in the handler collection** with `pipeline_file.check_type = PipelineFileCheckType.FORMAT_CHECK` (not full content, header only, with the assumption that if deeper validation is required, it will use a specific `CheckRunner` implementation for that purpose OR if the validation is not generic, it will be implemented by the specific handlers themselves)

aodncore/pipeline/steps/resolve.py
ocehugo: I think a valid and sane csv do not have trailing commas, unless you want to specify a empty column.

Maybe Option D: Just use a "single file per line" file!? Simple is better. This removes both the "another csvish formatted file to deal with" and the yagni/fields that I don't care problems.

I say that because creating these manifests is usually done via regexes/grep/cut pies or sql where queries. Thus, better to stick with a list of files only which is easily dumped straight away. If the need ever arises for specifying more than one thing for each , then we are better of creating a proper json/yaml manifest. 

Otherwise, just allow any csv input and pickup the columns that matches the columns on the database ('file_url",'publish/harvest_type',etc). At least this way we don't need reformatting and we can just feed in directly from a SQL query dump.

aodncore/pipeline/common.py
ocehugo: It's not overengineering to use a package to validate a CSV for a simple manifest file? Unless this is going to allow harvesting of generic CSV's into the database, I don't see reason why.

All that talk about csv reading, iterators and schema sounds very familiar! See https://github.com/aodn/python-aodndata/blob/7faf6a3a7e59913b948f62815ec84abba79f845a/aodndata/common/csv_schema.py.

Can't remember anymor, but from memory the code is a step above tableschema, which allow a generic function/class to work as a validation call in a per row basis. The performance is nothing something you could care if you are doing this though, mostly because of the function call overhead per row. I still blame the schema package for the performance hit, but it's python world afterall.

PS: Let me know if you guys prefer the tableschema thing, we shall update the SATTAG pipeline eventually to align with core.

PS2: Tableschema is not that flexible on the content validation procedure. Even the "schema" implementation can be finicking some times (even on standard csvs)... maybe it's better now, not sure. Also, the stacktrace when something fails was really bad.

lwgordonimos: > It's not overengineering to use a package to validate a CSV for a simple manifest file?

Including a high quality library vs writing code in house is *less* engineering! ;) 

The additional context you are probably missing is that we have (all but certainly) decided to use this library for another purpose (to interoperate with an external party using a common CSV schema, so don't want to be locked to Python), because it's language agnostic, uses JSON schema *documents* rather than just being Python code. It's also very standards based (following RFCs etc.) which is always good for long term maintainability.

> Let me know if you guys prefer the tableschema thing, we shall update the SATTAG pipeline eventually to align with core.

For the above reasons, we need the cross-platform, document-based schema definition approach either way (because we basically need to store the schema definitions as JSON, separately from any Python validation code), so once merged, you can use the library or not, depending on your use case.

ocehugo: @lwgordonimos , reading back today appears I'm talking about the code, but the overengineering was more about using a csv for input and all the luggage related to schema/validation/type-detection around it. So the point is not the code itself, but about using csv for the job.

My point is, or just go with a very simple file (e.g. the one file per line) that is cake to generate, or just go with a proper and robust schema file. If the input is too complex to by type by hand, then we will need to write the code anyway to create the manifest (and validate it, and raise errors about entries we care about, etc).

lwgordonimos: I agree it would be overkill to include the library for this use case alone, but once you already have it included, it's better to use it than write code. In terms of baggage, that only comes from lines of code having to be written that don't need to be written or unnecessary/lower quality library dependencies. 

Even in the simplest "one file per line" case, a one column CSV file which enforces unique values can be described and validated with *zero code*, using a library which we already include for other things e.g.

```python
{
  'name': 'dest_path',
  'type': 'string',
  'constraints': {
	'required': True,
	'unique': True
  }
}
```

Even the above simple definition does a lot of work that would otherwise have to be lines of Python code iterating over a file, testing for uniqueness etc. and we would *still* need to handle other CSV files anyway, which would mean two different ways of looping over and validating simple text files. 

Sometimes we need a pair of values coming through as a CSV as well (e.g. map manifests, with a source path and a destination path). So it's actually a good thing for simplicity and consistency if we can say "everything is a CSV, and gets handled through the same single process".

lbesnard (approved):

aodncore/pipeline/steps/resolve.py
mhidas: If I'm reading all this correctly, what @lwgordonimos means by "retain the current behaviour" and what @ocehugo means by "single file per line" are one and the same, and this is "Option A" [above](https://github.com/aodn/python-aodncore/pull/215#discussion_r508946484).

mhidas (approved):
I think it's good :+1:

aodncore/pipeline/common.py
ocehugo: Baggage is not the code to read/parse the input file, but all the other things you do because you can't validate the content at the parsing step. I'm talking about all the leakage into different classes names, with different schema properties, with different error handlings, and general handling because you are limited to the validation grammar of the library.

I'm by all means on the reuse what is currently being used, and I understand the above case of picking a library that is common case across the board. However, I only see a partial solution. For example, using `schema` gives you the power and the nice declarative things.
simple?
```
delete_schema = {'dest_path': string, required=['dest_path']};
Schema(delete_schema).validate(input)
```

complex?
```python
import complex_dest_path_validation
schema = {'dest_path': complex_dest_path_validation, required=['dest_path']};
#`is_unique` could be anything, even a complex class that do the validation or other things.
```

`schema` substitute all validation libaries that I know of (and goes really nice with data classes if complexity is needed), reduce test verbosity, class methods and a lot of the boiler plate. Just compare a functional style schema set of validation functions to an equivalent class/object oriented validation procedure.

PR #220, opened 2020-12-15
Add 'allow_delete_manifests' flag to resolve_params

lwgordonimos:
This feature requires that a pipeline(or handler) explicitly allows delete_manifests before resolving them, by setting the flag in the `resolve_params` dictionary.


mhidas (approved):
:+1:

PR #222, opened 2020-12-17
Update validate_membership to make testing Enums work consistently ac…

lwgordonimos:
…ross Python versions 3.5-3.8 by raising a TypeError when testing for non-Enum members


lwgordonimos:
This should be mergeable, and is to "future-proof" for Python 3.8 support.

PR #223, opened 2021-01-11
Enum type error

lwgordonimos:


lwgordonimos:
This should be mergeable, and is to "future-proof" for Python 3.8 support.

PR #234, opened 2021-06-24
Generic CSV Harvester - DRAFT PR

ndhogan:
For https://github.com/aodn/backlog/issues/3149

Still to do:
- [ ] finalise the pipeline location and/or environment variable for public-schema and db connection strings - these are dependent on outstanding chef changes
- [ ] build the tableschema validation in check step - future iteration
- [ ] build the geonetwork extents update function - future iteration

PR #235, opened 2021-06-30
Remove numpy version limit in gh actions

mhidas:
Don't know why this was limited in the first place, but it's now causing errors, so let's get rid of it!

From tests in virtualenvs on my machine, changing the numpy version alone seems to cause or fix the "numpy.ndarray size changed, may indicate binary incompatibility" errors.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -25,7 +25,6 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        pip install "numpy<1.19.0"
         pip install -r test_requirements.txt
         pip install pytest-cov
     - name: Test with pytest
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [ '3.5.2', '3.8' ]
+        python-version: [ '3.5', '3.8' ]
     steps:
     - uses: actions/checkout@v2
     - name: Install libudunits2-dev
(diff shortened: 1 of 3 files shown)


ndhogan:
Hmm... ok, so this is producing inconsistent results.  On my local machine ```pip install -r test_requirements``` will succeed (cf-units==2.0.2) without requiring _numpy_ to be pre-installed, whereas in the github action it fails on downloading cf-units.

Compare with https://github.com/aodn/python-aodncore/pull/234 where the github action still pre-installs numpy<1.19.0 and passes on [3.5](https://github.com/aodn/python-aodncore/runs/2947890485?check_suite_focus=true), but then fails on [3.8](https://github.com/aodn/python-aodncore/runs/2947890499?check_suite_focus=true) (seemingly because 3.8 requires numpy>1.19.0 - on my local machine 3.8 installs numpy==1.21.0 and all tests pass)

ndhogan:
ok, so the upshot is:

1. Production and RC are running Python 3.5.2, but github actions can only generate an environment for Python 3.5.10 (in the 3.5 minor release) - however, the difference only seems to be an issue when using cached packages on a local machine
2. Python 3.5.x requires that numpy<1.19.0 be pre-installed for an appropriate cf-units package to be downloaded, otherwise the _gitub actions-->Install dependencies_ step will fail
3. Python 3.8 does not require numpy to be pre-installed, but the _github actions-->Test with pytest_ step will fail if numpy<1.19.0 __is__ installed
4. Unittest and/or pytest will sometime fail on a local machine for Python 3.5.x if the virtual environment uses the stored cache to install dependencies, so for consistent results:
```bash
python -m pip install --upgrade pip # in Python3.5.2 pin version ==20.3.4 as later versions will introduce error string formatting that causes a command error
pip cache purge # this will ensure that "pip install" won't use the stored cache
pip install numpy
pip install -r test_requirements.txt
pytest
```

I think that until we upgrade to Python 3.8, we will need to continue pre-installing numpy, but drop the <1.19.0 constraint so that Python 3.5.x can continue to install numpy==1.18.5 while Python 3.8.x can install a later (compatible) version.

PR #242, opened 2021-11-01
Fixes for new WSD code

ndhogan:
Relates to:
* https://github.com/aodn/python-aodncore/pull/237#discussion_r737127389
* https://github.com/aodn/python-aodncore/pull/237/files#r737127389
* https://github.com/aodn/python-aodncore/pull/237/files#r687439496


mhidas:
I'll have a look now - but noticed that GitHub Actions also didn't get triggered for this PR @ndhogan

mhidas (approved):
That all looks good to me.
My only comment is that it doesn't look like the added file "test_frictionless_no_resource.csv"  is used in any test??
Even if it was, you could truncate it to just the first 10 lines?

aodncore/pipeline/steps/harvest.py
```diff
@@ -435,7 +436,13 @@ def run(self, pipeline_files):
         self.pipeline_files = pipeline_files
         objs = self.params.get('db_objects')
         if objs:
-            runsheet = (x for x in map(self.build_runsheet, objs) if x)
+            runsheet = [x for x in map(self.build_runsheet, objs) if x]
```
ndhogan, line 439: changed _runsheet_ from generator to list

mhidas: :+1: Thanks. That'll do it!

mhidas (approved):
:+1: :+1: I've tested this in the pipeline box and all seems to run as expected now!

The only last thing is I still don't think we need the file test_aodncore/testdata/test_frictionless_no_resource.csv

mhidas:
> I still don't think we need the file test_aodncore/testdata/test_frictionless_no_resource.csv

Actually, I was wrong. It's used in test_check.py  However its content is irrelevant because it's just a file with no matching schema to validate it. So I've truncated it to the first 10 lines just to trim a bit of fat off the repo.  :smirk:

PR #245, opened 2021-11-30
use pyenv in docker build

digorgonzola:


mhidas (commented):
I don't know much about the ins & outs of docker, so I'm probably not the right person to review this. Maybe @ndhogan has some comments?

digorgonzola:
> 

Just included you so you would be aware it was a possibility. It would allow us to have python projects build against specific python versions without running legacy docker images. Out of interest, there is also the ability to use pyenv in github actions which gives you better control of python versions rather than the limited versions available currently.

mhidas:
Thanks @digorgonzola 
> Out of interest, there is also the ability to use pyenv in github actions which gives you better control of python versions rather than the limited versions available currently.

That sounds useful, especially if we could somehow leverage the same config used to set up the pyenv environment in RC/prod to ensure we're testing in the exact same environment in GH Actions. Could we use the same Docker image built by Jenkins?

Dockerfile
```diff
@@ -1,38 +1,66 @@
...
     ca-certificates \
     git \
     libmagic1 \
     libudunits2-dev \
     python3-dev \
     wget \
-    # docker.io \
+    libffi-dev \
+    # Pyenv pre-requisites
+    make zlib1g-dev libbz2-dev libreadline-dev libsqlite3-dev wget curl llvm libncurses5-dev libncursesw5-dev xz-utils tk-dev libffi-dev liblzma-dev python-openssl git \
```
akashisama, line 34: This set of packages fails for installing other versions of python, eg 3.8.12. Full set of recommended dependencies at https://github.com/pyenv/pyenv/wiki#suggested-build-environment

digorgonzola: I don't seem to be hitting any such problem... Can you share the exact error?

I've tested this with 3.5.2, 3.8.10, 3.8.12. Later versions fail due to python dependency issues but I'm not going to waste any time on that now :)

Dockerfile
```diff
@@ -1,38 +1,68 @@
-FROM ubuntu:16.04
+FROM ubuntu:latest
```
mhidas, line 1: Should this be explicitly `ubuntu:20.04`? Otherwise won't `latest` automatically become `22.04` at some point in the not too distant future?

akashisama (approved):

Dockerfile
digorgonzola: Yep! However, that would be fine since we are not reliant on the system version of python. It just means our build environment will be up-to-date and secure. 

It may mean a quick revisit to ensure that pyenv and any python dependencies don't break but we should be able to fix this quickly. If not, a quick change to fix the docker container back to ubuntu 20.04

mhidas: :+1:

PR #248, opened 2022-01-31
Stop jenkins from attempting to bump version on non-master branches

akashisama:


akashisama:
Changes incorporated into https://github.com/aodn/python-aodncore/pull/245

PR #250, opened 2022-02-08
celery must be less than version 5

digorgonzola:
celery version 5 introduces new behaviour that is not compatible with our existing watch config

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -8,7 +8,7 @@ ENTRY_POINTS = {
 
 INSTALL_REQUIRES = [
     'boto3>=1.9.156',
-    'celery>=4.3.0',
+    'celery>=4.3.0,<5',
     'compliance-checker==4.1.1',
     'jsonschema>=2.6.0',
     'paramiko>=2.6.0',


digorgonzola:
Celery version 5+ will raise the following error:

``` bash
Error: no such option: --app
You are using `--app` as an option of the worker sub-command:
celery worker --app celeryapp <...>

The support for this usage was removed in Celery 5.0. Instead you should use `--app` as a global option:
celery --app celeryapp worker <...>
Usage: celery worker [OPTIONS]
Try 'celery worker --help' for help.
```

mhidas (commented):
Looks pretty straightforward. I just wonder if it would make more sense to put the `<5` limit in the [constraints.txt](https://github.com/aodn/python-aodncore/blob/bf59e94304ae7ee12b799f4832318950422b2225/constraints.txt) file instead?

digorgonzola:
> Looks pretty straightforward. I just wonder if it would make more sense to put the `<5` limit in the [constraints.txt](https://github.com/aodn/python-aodncore/blob/bf59e94304ae7ee12b799f4832318950422b2225/constraints.txt) file instead?

Ah sure! Happy with that approach too. I'll force push a new commit.

mhidas:
It might be worth considering whether it's any code in this repo itself that doesn't work with celery 5?
Or is the issue with configuration that lives (and could potentially be updated) elswhere, without any changes in aodncore? In that case maybe the constraint shouldn't be in this repo at all?

PR #251, opened 2022-02-21
set package version on master branch to dev

digorgonzola:
this is to allow dev testing in vagrant without it pulling in the packaged version from PyPI


akashisama (changes requested):
we'll need to think about this a bit more

PR #252, opened 2022-03-15
Update README.md

craigrose:


README.md
```diff
@@ -1,4 +1,6 @@
-## python-aodncore
+## AODN Core Pipeline
+=
+The aodncore.pipeline package provides the base class for each pipeline handler, aodncore.pipeline.HandlerBase. This is the starting point for any new handler development, as it contains all of the core functionality of each handler, which is then available to the child class via class inheritance.
```
mhidas, line 3: Just adding my 2c's worth... If the goal is to provide a simple high-level summary, wouldn't it be best to avoid implementation details like "handler", and define what sort of "pipelines" we're talking about? How about...
> The `aodncore` package provides the core & generic functionality for all data ingestion pipelines at the AODN. This can be customised and extended for specific data-streams in the [`aodndata`](https://github.com/aodn/python-aodndata) package.

(also, I don't know why there's a '=' on line 2?)

digorgonzola (approved):

PR #257, opened 2022-07-01
Refactor SMTP client initlising method and update unit tests

vietnguyengit:


vietnguyengit:
@digorgonzola No luck yet to setup the pipeline vagrant node, so I cannot execute the actual code, but so far Unit Tests all passed as well as Codecov checks in this new PR, it will be good if you can pull this branch and test on your vagrant node to see how it will go.

My attempt to run `vagrant up ...` always get this error:
```
An error occurred in the underlying SSH library that Vagrant uses.
The error message is shown below. In many cases, errors from this
library are caused by ssh-agent issues. Try disabling your SSH
agent or removing some keys and try again.

If the problem persists, please report a bug to the net-ssh project.

timeout during server version negotiating
```

After trying different way to initlise the `smtp_server`, I think rather than let `_send()` method to do everything, delegating the job to a static method `smtp_server_init()` will make it's more organised, avoid duplicated `connect()` invoke, and the Unit tests make sense: with no server connection - nothing can happen properly after that.


I'll change this PR to ready once tests on the vagrant node are successful.

vietnguyengit:
I'm now able to ssh into the vagrant pipeline prod node after wiping out the existing VirtualBox VMs and run `vagrant up ...` again but have no clue what to do here to test the `notify.py` functions. Will need your help on Monday to walk me through the processes with vagrant node @digorgonzola Have a nice weekend.

digorgonzola (approved):

PR #261, opened 2022-07-08
Potential way to do versioning based on tags 

alexgleith:
Creating a draft PR for discussion.

This will make it possible to do a local install the "normal" way and get a version number, which is required for the three `python-aodnX` packages to interact with each other. Manually patching the setup.py file is not ideal.

I don't know the impact on the CI process later down the chain, but this _should_ mean that the `bumpversion.sh` can just be removed and the package will install happily.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
+# Ignore the version file, it's created on install
+aodncore/_version.py
+
 # Byte-compiled / optimized / DLL files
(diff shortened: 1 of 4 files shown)


alexgleith:
Closing because @digorgonzola did a better job here: https://github.com/aodn/python-aodncore/pull/263/files

PR #264, opened 2022-09-06
4328 setuptools handle runtime version

digorgonzola:


vietnguyengit (dismissed):

vietnguyengit:
There is a failed test related to module import I noticed.

mhidas:
> There is a failed test related to module import I noticed.

It's failing on 
```
aodncore/pipeline/handlerbase.py:24: in <module>
    from .._version import __version__ as _aodncore_version
E   ImportError: cannot import name '__version__'
```
... but only in python 3.5, which we're no longer using. We should remove that version from the [test config](https://github.com/aodn/python-aodncore/blob/da9c9d642b8f649e67f00da6ba77913c0c162b0f/.github/workflows/test.yml#L15)

I'm not sure I understand why there's a difference - I guess the handling of relative imports has changed between 3.5 and 3.8?

aodncore/pipeline/handlerbase.py
```diff
@@ -21,7 +21,7 @@
 from ..util import (ensure_regex_list, ensure_writeonceordereddict, format_exception,
                     get_file_checksum, iter_public_attributes, lazyproperty, matches_regexes, merge_dicts,
                     validate_relative_path_attr, TemporaryDirectory, WfsBroker, DEFAULT_WFS_VERSION)
-from ..version import __version__ as _aodncore_version
+from .._version import __version__ as _aodncore_version
```
mhidas, line 24: Would it be cleaner to just import it from the package root? (where we've already [handled the change](https://github.com/aodn/python-aodncore/blob/master/aodncore/__init__.py) in version source?)
```suggestion
from aodncore import __version__ as _aodncore_version
```

sphinx/conf.py
```diff
@@ -20,7 +20,7 @@
 import sys
 
 sys.path.insert(0, os.path.abspath('..'))
-import aodncore.version
+import aodncore._version
```
mhidas, line 23: Similarly here...
```suggestion
from aodncore import __version__
```
... and update below accordingly.

vietnguyengit:
@digorgonzola  There is still a failed test, but this time with `codecov`: https://github.com/aodn/python-aodncore/pull/264/checks?check_run_id=8223495289

vietnguyengit (approved):

PR #266, opened 2022-09-09
4290 prefect pipeline

craigrose:
These modify aodncore to allow for running of pipelines with Prefect and S3 buckets. The changes should not affect existing pipeline operations.


aodncore/util/s3_util.py
```diff
@@ -0,0 +1,64 @@
+from urllib.parse import urlparse
+import boto3
```
mhidas, line 2: Some functionality for handling objects on S3 is already implemented as an [`S3StorageBroker`](https://github.com/aodn/python-aodncore/blob/0fe87c95681ae535fd6eabcae352cb806dcfb0f7/aodncore/pipeline/storage.py#L304) class that suports query, download, upload and delete operations. Could that be reused instead of duplicating the functionality here?

aodncore/pipeline/prefecthandlerbase.py
```diff
@@ -0,0 +1,60 @@
...
+from aodncore.util import ensure_regex_list
+
+
+class PrefectHandlerBase(HandlerBase):
+
+    def _set_input_file_attributes(self):
+        """ Override HandlerBase"""
+
+        try:
+            self._file_checksum = self.etag
```
mhidas, line 17: I can't see where this `etag` attribute is set?

mhidas (commented):
Just adding my 2c's worth while I look at this.

PR #267, opened 2022-09-21
TableSchema check improvements

mhidas:
Trying to improve and slightly extend the functionality for this check type to support the Animal Tracking use case.


digorgonzola (approved):

test_aodncore/pipeline/steps/test_check.py
```diff
@@ -223,3 +235,16 @@ def test_missing_schema_file(self):
...
+                                           check_params={"tableschema_filename_pattern": "[^_]*(_[^_]*)?"}
+                                           )
+        collection = PipelineFileCollection(SAME_SCHEMA_CSVS)
+        ts_runner.run(collection)
+
+        failed_files = [(f.name, f.check_result.log)
+                        for f in collection
+                        if not f.check_result.compliant
+                        ]
+        self.assertEqual([], failed_files)
```
alexgleith, line 250: You could do a length check, instead of comparing two empty tuples.

alexgleith (approved):
Looks good to me.

One improvement might be to add a test that finds a file that doesn't match a pattern, so test for a failed check.

aodncore/pipeline/schema.py
```diff
@@ -22,7 +22,8 @@
         'criteria': {'type': 'string'},
         'skip_checks': {'type': 'array', 'items': {'type': 'string'}},
         'output_format': {'type': 'string'},
-        'verbosity': {'type': 'integer'}
+        'verbosity': {'type': 'integer'},
+        'tableschema_filename_pattern': {'type': 'array'}
```
evacougnon, line 26:
```suggestion
        'tableschema_filename_pattern': {'type': 'string'}
```

test_aodncore/pipeline/steps/test_check.py
mhidas: :+1: Actually I didn't even need to create a list (it was just helpful while debugging)

mhidas:
@digorgonzola @nspool This is ready to merge, if you have a few minutes for a final look. I've added a bit more to the unittests after suggestions from @alexgleith .

nspool:
When I run the tests locally the test `test_compliant_file` is failing for me with the warning:
```
DeprecationWarning: Converting `np.character` to a dtype is deprecated. The current result is `np.dtype(np.str_)` which is not strictly correct. Note that `np.character` is generally deprecated and 'S1' should be used.
```

Running Python 3.8.16

nspool:
@mhidas after running `pip install -r test_requirements.txt` these versions are installed:

```
compliance-checker        4.1.1
numpy                     1.24.2
```

Trying `pip install compliance-checker==5.0.1` fails as that version is not available, only `5.0.0` or `5.0.2`.

nspool:
Thanks @mhidas , I rebased from master locally and all the tests pass. Happy to merge.

nspool (approved):

mhidas:
:+1: Thanks @nspool

PR #268, opened 2022-09-27
watch service ignore non-existent directories

gsatimos:
Backlog: https://github.com/aodn/backlog/issues/4308

In case there is a misconfigured directory (as per this issue: https://github.com/aodn/issues/issues/1257) this prevents the watch service from failing to start by just skipping the directory and logging a warning which could then be picked up elsewhere e.g. sumo logic

### DIFF
--- a/aodncore/pipeline/watch.py
+++ b/aodncore/pipeline/watch.py
@@ -50,7 +50,7 @@ from celery import Task
 from celery.utils.log import get_task_logger
 
 from .exceptions import InvalidHandlerError
-from ..util import list_regular_files, safe_move_file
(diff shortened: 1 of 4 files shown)


gsatimos:
draft until I work out how to execute tests properly on my local

gsatimos:
@sqbaillie the above `codecov/patch` and `codecov/project` are complaining about a reduction in test coverage for a section of code that didn't have any tests written at all. What happens now? ignore them or do I need to implement a heap of tests that have nothing to do with the code I updated?

gsatimos:
ping @sqbaillie or maybe @digorgonzola regarding above comment... what should I do?

digorgonzola (approved):

PR #271, opened 2023-01-08
1306 fix circular import

utas-raymondng:
test_gsla in python-aodndata shows circular dependency problem before the fix, right now the circular dependency fixed but show other error which didn't seems to be issue cause by this fix.

# Summary of fix
1. Refactor ff.py so that it will not cause circular dependencies.
2. Upgrade compliance-checker to v5.0.2 because it contains a fix to numpy.int (which no longer support as of 1.24.0) 
3. Force the numpy to below 1.24.0 because compliance-checker 5.0.2 still have problem with numpy 1.24.0, it call numpy.float which no longer support.

Item 2 is optional upgrade as we forced the version to below 1.24.0, but we need to upgrade anyway later when we can use numpy 1.24.0 or above.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -28,6 +28,7 @@ jobs:
         pip install numpy
         pip install -r test_requirements.txt
         pip install pytest-cov
+        pip list
     - name: Test with pytest
       run: |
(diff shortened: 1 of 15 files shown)


utas-raymondng:
Do not merged this

PR #272, opened 2023-01-09
Compliance-checker use obsoleted numpy.float, use 1.23.5 numpy fixed …

utas-raymondng:
Force numpy version as it is having problem with compliance-checker where it cause issue during import

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -17,6 +17,9 @@ INSTALL_REQUIRES = [
     'tabulate>=0.8.2',
     'transitions>=0.7.1',
     'psycopg2-binary==2.8.6',
+    # Compliance-checker 5.0.1 have problem with numpy.float / numpy.int
+    # Starting 1.24.0 numpy.float / numpy.int no longer support and should use float / int instead.
+    'numpy<1.24.0',
     'PyYAML==5.3.1'
 ]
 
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -17,6 +17,9 @@
     'tabulate>=0.8.2',
     'transitions>=0.7.1',
     'psycopg2-binary==2.8.6',
+    # Compliance-checker 5.0.1 have problem with numpy.float / numpy.int
+    # Starting 1.24.0 numpy.float / numpy.int no longer support and should use float / int instead.
+    'numpy<1.24.0',
```
mhidas, line 22: Ok, but we have a separate [constraints.txt](https://github.com/aodn/python-aodncore/blob/master/constraints.txt) file for such conditions, so this should really go in there. Those constraints are automatically included by commands like `pip install -r requirements.txt`

mhidas:
Might also be worth update the GH Action workflow file? For some reason it explicitly installs numpy before `pip install -r test_requirements.txt`. I can't remember why that was necessary -- maybe it isn't anymore?
https://github.com/aodn/python-aodncore/blob/fbd0175c13f59f92d038093485c8c04accb536e6/.github/workflows/test.yml#L28

mhidas:
Note that IOOS have fixed this issue on their master branch (https://github.com/ioos/compliance-checker/pull/944) so the next time we're able to upgrade to a newer release, this constraint can be removed.

utas-raymondng:
@mhidas updated a per comments

mhidas (approved):

